PR #38, opened 2023-02-09
feat: all resources are private by default

SimonMarquis:


soulcramer (approved):
Nice, I forgot to create an issue for this

PR #42, opened 2023-02-10
feat(git): prepare LFS support for snapshots testing

### LINKED ISSUE #22: Enable Git LFS
We will need this to have better screenshot testing with paparazzi for #7 
We used these to inspire ourselves for these tests:
- https://github.com/DroidKaigi/conference-app-2022/pull/97/files
- https://github.com/slackhq/circuit/pull/204/files

SimonMarquis:
Closes #22

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+spark/**/snapshots/**/*.png filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+spark/**/snapshots/**/*.png filter=lfs diff=lfs merge=lfs -text
```
SimonMarquis, line 1:
```suggestion
**/src/test/snapshots/**/*.png filter=lfs diff=lfs merge=lfs -text
```

SimonMarquis:
@soulcramer is it OK for you if we merge this filter now, and prepare workflows later?

soulcramer (approved):
Yeah sure 👍

PR #50, opened 2023-02-10
fix: data objects requires language version 1.9

SimonMarquis:


soulcramer (approved):
😢 
Do you know whats blocking us from using the language version 1.9?

SimonMarquis:
We need to configure it until it become stable

```kotlin
tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
    kotlinOptions.languageVersion = "1.9"
}
```

And this would force consumer to update as well, otherwise they will get an error like this:
```
Class 'com.adevinta.spark.Spark' is compiled by a pre-release version of Kotlin and cannot be loaded by this version of the compiler
```

soulcramer:
Ah Right! Thanks for the explanation 👍

soulcramer (approved):

PR #52, opened 2023-02-10
feat(workflow): add LFS validation to the validation job

SimonMarquis:


SimonMarquis:
@soulcramer It seems like it worked like a charm 🍀
Doo we merge this in the codebase?

soulcramer (approved):
Yes let's do it 😎

PR #58, opened 2023-02-13
feat(pr): comment on PR with error message

SimonMarquis:


SimonMarquis:
The message posted by the bot now comes directly from the `cz check` output.

soulcramer (approved):

PR #81, opened 2023-02-15
feat: enable explicitApi for Android module

SimonMarquis:
Must be configured with `freeCompilerArgs`...


soulcramer (dismissed):

SimonMarquis:
@soulcramer this can be a bit tedious because it needs to validate test sources as well.
But we have an immediate feedback on the IDE (and auto-fix) so 🤷 
image

soulcramer (approved):
Arf, looks like it will be tedious at first

SimonMarquis:
During the "import" phase, I agree. But this will be beneficial for sure.

PR #83, opened 2023-02-15
fix: wrong Kotlin version introduced by mistake

SimonMarquis:


spark/build.gradle.kts
```diff
@@ -55,7 +55,7 @@ android {
         allWarningsAsErrors = true
         freeCompilerArgs += listOf(
             "-Xexplicit-api=strict",
-            "-P", "plugin:androidx.compose.compiler.plugins.kotlin:suppressKotlinVersionCompatibilityCheck=1.8.20-Beta",
```
soulcramer, line 58: We need to remove it entirely if the version is the correct one for our compiler version otherwise it'll also print an error (yeah it's dumb)

soulcramer (changes requested):

spark/build.gradle.kts
SimonMarquis: Oh, I see. Then is it really necessary for this project at all?

soulcramer: I think so since if the kotlin version that we use in the project in not the same as the kotlin version compatible with the Compose Compiler version then we'll need to specify it or we'll get a warning/error :hidethepain:

soulcramer (approved):

PR #85, opened 2023-02-15
feat: Add issue templates

soulcramer:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,50 @@
...
+      options:
+        - label: I have searched the existing issues
+          required: true
+  - type: dropdown
+    attributes:
+      label: What version of spark are you using?
+      description: This will help us identify where & when the bug may have appeared
+      options:
+        - 1.0.0
+        - Other (please specify it)
```
SimonMarquis, line 23: Is there a `required: true` available for this or do we want to allow users to report issue that are not tied to a version?

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,46 @@
+name: Feature request
+description: File a feature request
+title: "[FR]: "
+labels: ["enhancement"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+       Thanks for taking the time to fill out this bug report!
```
SimonMarquis, line 9: This should not be there in a feature request template imo

.github/ISSUE_TEMPLATE/bug_report.yml
soulcramer: Yep, Ì'll add it as a requirement

SimonMarquis (approved):

PR #119, opened 2023-02-20
test: Add Local screenshot testings from Compose Previews

soulcramer:
## 🧑‍ Changes description

- Add screenshots-test module that will depend on every other spark component modules
- We disable the CI test on `screenshots-test` as it only run on debug variant since we use debug dependencies from other modules

## 🤔 Context

This allow us to have a dedicated modules for screenshot testing and avoid us from implementing it on every modules instead.

## 📸 Screenshots

🤩
com adevinta spark_PreviewScreenshotTests_preview_tests_toggles_checkbox_part_light

## 👋 Other info

Start of #7 but we still need to add a CI test for screenshot testing to block it on failure.


gradle/libs.versions.toml
```diff
@@ -44,6 +46,9 @@ androidx-compose-ui-tooling = { module = "androidx.compose.ui:ui-tooling" }
 androidx-compose-ui-tooling-preview = { module = "androidx.compose.ui:ui-tooling-preview" }
 androidx-compose-ui-util = { module = "androidx.compose.ui:ui-util" }
 
+# https://github.com/google/TestParameterInjector
+google-testParameterInjector = { module = "com.google.testparameterinjector:test-parameter-injector", version.ref = "testParameterInjector" }
```
SimonMarquis, line 50: IMO we should not prefix everything as it makes it harder to find what you look for.
I think we should keep `testParameterInjector` and `paparazzi`

.github/workflows/ci.yml
```diff
@@ -68,7 +68,7 @@ jobs:
         id: test
         uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef # v2.3.3
         with:
-          arguments: testReleaseUnitTest
+          arguments: testReleaseUnitTest -x :screenshots-test:testReleaseUnitTest
```
SimonMarquis, line 71: This is error prone. If we don't want to run "release" tests of this module, we should simply disable this buildType entirely.

screenshots-test/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,26 @@
...
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+  SOFTWARE.
+  -->
+<manifest>
+
+</manifest>
```
SimonMarquis, line 26: Is it really necessary?
If yes, collapse the empty tag:
```suggestion
<manifest />
```

screenshots-test/src/test/kotlin/com/adevinta/spark/PaparazziUtils.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.compose.foundation.layout.Box
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalInspectionMode
+import androidx.compose.ui.unit.Dp
+import app.cash.paparazzi.Paparazzi
+
+internal fun Paparazzi.snapshotPlus(
```
SimonMarquis, line 36: nit: Can we rename this into sparkSnapshot instead? (the "plus" does not bring valuable info)

```suggestion
internal fun Paparazzi.sparkSnapshot(
```

screenshots-test/src/test/kotlin/com/adevinta/spark/PreviewScreenshotTests.kt
```diff
@@ -0,0 +1,84 @@
...
+import app.cash.paparazzi.Paparazzi
+import com.airbnb.android.showkase.models.Showkase
+import com.airbnb.android.showkase.models.ShowkaseBrowserComponent
+import com.android.ide.common.rendering.api.SessionParams
+import com.google.testing.junit.testparameterinjector.TestParameterInjector
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(TestParameterInjector::class)
```
SimonMarquis, line 35: Why are we using `TestParameterInjector` if there is no `@Parameter` in the tests?

screenshots-test/src/test/kotlin/com/adevinta/spark/PreviewScreenshotTests.kt
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    @Test
+    fun preview_tests() {
+        Showkase.getMetadata().componentList.groupBy {
+            // Currently Compose Previews generated by PreviewParameters have the same name.
+            // So we need to group them to later manually add their variants
+            previewFullName(it)
+        }.forEach { (name, components) ->
+            components.forEachIndexed { index, showkaseComponent ->
```
SimonMarquis, line 64: Shouldn't this be generated by the `TestParameterInjector`?

screenshots-test/src/test/kotlin/com/adevinta/spark/PreviewScreenshotTests.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.airbnb.android.showkase.models.Showkase
+import com.airbnb.android.showkase.models.ShowkaseBrowserComponent
+import com.android.ide.common.rendering.api.SessionParams
+import com.google.testing.junit.testparameterinjector.TestParameterInjector
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(TestParameterInjector::class)
+internal class PreviewScreenshotTests {
```
SimonMarquis, line 36: This file should probably be in the debug sourceSet as we depend on Showkase, which is added to the debug classpath.

screenshots-test/src/test/kotlin/com/adevinta/spark/PreviewScreenshotTests.kt
```diff
@@ -0,0 +1,84 @@
...
+                paparazzi.snapshotPlus(name = previewNameWithVariants) {
+                    showkaseComponent.component()
+                }
+            }
+        }
+    }
+}
+
+private val partPro = listOf("_part_light", "_pro_light", "_part_dark", "_pro_dark")
+private val lightDark = listOf("_light", "_dark")
```
SimonMarquis, line 81: This feels clunky and error prone. Is there a way we can reference the right properties from the preview parameters instead?

settings.gradle.kts
```diff
@@ -45,3 +45,4 @@ rootProject.name = "spark-android"
 include(":spark")
 include(":sample")
 include(":spark-lint")
+include(":screenshots-test")
```
SimonMarquis, line 48: nit: I'm not fond of this naming. I was thinking of using a common prefix for all spark modules. To me, `spark-screenshot-testing` would feel more natural. WDYT?

screenshots-test/src/main/AndroidManifest.xml
soulcramer: It isn't! Will collapse it

SimonMarquis: Could be removed entirely

screenshots-test/src/test/kotlin/com/adevinta/spark/PreviewScreenshotTests.kt
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+package com.adevinta.spark
+
+import app.cash.paparazzi.DeviceConfig
+import app.cash.paparazzi.Paparazzi
+import com.airbnb.android.showkase.models.Showkase
+import com.airbnb.android.showkase.models.ShowkaseBrowserComponent
+import com.android.ide.common.rendering.api.SessionParams
+import com.google.testing.junit.testparameterinjector.TestParameterInjector
```
SimonMarquis, line 30:
```suggestion
```

screenshots-test/src/test/kotlin/com/adevinta/spark/PreviewScreenshotTests.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
```
SimonMarquis, line 33:
```suggestion
```

gradle/libs.versions.toml
```diff
@@ -63,6 +68,7 @@ android-kotlin = { id = "org.jetbrains.kotlin.android", version.ref = "kotlin" }
 android-library = { id = "com.android.library", version.ref = "android-gradlePlugin" }
 android-lint = { id = "com.android.lint", version.ref = "android-gradlePlugin" }
 android-test = { id = "com.android.test", version.ref = "android-gradlePlugin" }
+cashapp-paparazzi = { id = "app.cash.paparazzi", version.ref = "paparazzi" }
```
SimonMarquis, line 71:
```suggestion
paparazzi = { id = "app.cash.paparazzi", version.ref = "paparazzi" }
```

spark-screenshot-testing/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
```
SimonMarquis, line 1: nit: I think this file can be removed

spark/build.gradle.kts
```diff
@@ -149,11 +149,11 @@ dependencies {
...
 
-    debugImplementation(libs.airbnb.showkase)
-    kspDebug(libs.airbnb.showkase.processor)
+    implementation(libs.androidx.compose.ui.testManifest)
+    implementation(libs.androidx.compose.ui.tooling)
 
-    testImplementation(libs.junit)
-    testImplementation(libs.kotlin.test)
+    implementation(libs.airbnb.showkase)
+    ksp(libs.airbnb.showkase.processor)
```
SimonMarquis, line 156: We will have to triple check this in the near future, 'cause we really don't want to ship "debug" things in the release artifact of spark.

spark/src/main/kotlin/com/adevinta/spark/components/toggles/CheckBox.kt
```diff
@@ -73,7 +73,8 @@ public fun Checkbox(
         interactionSource = interactionSource,
         enabled = enabled,
         modifier = modifier,
-    )
+
+        )
```
SimonMarquis, line 77:
```suggestion
    )
```

SimonMarquis (approved):

build.gradle.kts
```diff
@@ -27,4 +27,5 @@ plugins {
     alias(libs.plugins.android.kotlin) apply false
     alias(libs.plugins.kotlin.jvm) apply false
     alias(libs.plugins.google.ksp) apply false
+    alias(libs.plugins.cashapp.paparazzi) apply false
```
SimonMarquis, line 30:
```suggestion
    alias(libs.plugins.paparazzi) apply false
```

spark-screenshot-testing/build.gradle.kts
```diff
@@ -0,0 +1,83 @@
...
+ * SOFTWARE.
+ */
+
+// TODO: Remove once https://youtrack.jetbrains.com/issue/KTIJ-19369 is fixed
+@Suppress("DSL_SCOPE_VIOLATION")
+plugins {
+    alias(libs.plugins.android.library)
+    alias(libs.plugins.android.kotlin)
+    alias(libs.plugins.google.ksp)
+    alias(libs.plugins.cashapp.paparazzi)
```
SimonMarquis, line 29:
```suggestion
    alias(libs.plugins.paparazzi)
```

settings.gradle.kts
soulcramer: Done in ca348372

.github/workflows/ci.yml
soulcramer: Done in a792de44

gradle/libs.versions.toml
soulcramer: Done in a792de44

screenshots-test/src/test/kotlin/com/adevinta/spark/PreviewScreenshotTests.kt
soulcramer: Done in a792de44

screenshots-test/src/test/kotlin/com/adevinta/spark/PreviewScreenshotTests.kt
soulcramer: Done in a792de44

SimonMarquis (approved):

spark/build.gradle.kts
```diff
@@ -149,11 +149,11 @@ dependencies {
     debugImplementation(libs.androidx.compose.ui.testManifest)
     debugImplementation(libs.androidx.compose.ui.tooling)
 
-    debugImplementation(libs.airbnb.showkase)
-    kspDebug(libs.airbnb.showkase.processor)
+    implementation(libs.androidx.compose.ui.testManifest)
+    implementation(libs.androidx.compose.ui.tooling)
 
-    testImplementation(libs.junit)
-    testImplementation(libs.kotlin.test)
```
SimonMarquis, line 156: We should probably keep those.

spark-screenshot-testing/src/test/kotlin/com/adevinta/spark/PreviewScreenshotTests.kt
```diff
@@ -67,10 +67,11 @@ internal class PreviewScreenshotTests {
                         else -> name
                     }
                     paparazzi.unsafeUpdateConfig(deviceConfig = BaseDeviceConfig.PIXEL_5.deviceConfig)
-                    paparazzi.snapshotPlus(name = previewNameWithVariants) {
+                    paparazzi.sparkSnapshot(name = previewNameWithVariants) {
                         showkaseComponent.component()
                     }
-                } catch (e: Exception) {
+                } catch (t: Throwable) {
+                    // TODO-@soulcramer (22-48-2023): Should we display the throwable message here too?
```
SimonMarquis, line 74: 😬 What I told you the other day was in fact wrong.
If we keep these tests in a single `@Test` method we won't be able to make any use of the multiple potential errors...

SimonMarquis: Let's ignore this for now until we have dedicated test methods for each composable. WDYT?

soulcramer: Yeah I know but we said that it'll at least until have dedicated components tests to avoid the test to fail on the first screenshot that is not the same.

SimonMarquis (approved):

spark-screenshot-testing/src/test/kotlin/com/adevinta/spark/PreviewScreenshotTests.kt
SimonMarquis: Oh I see. The images will be in the artifact, even though there is only a single test. 👍

PR #124, opened 2023-02-21
feat: paparazzi tests and reports

### LINKED ISSUE #7: Setup the CI to run screenshot testing
We just want it to run like it did on the LBC app

SimonMarquis:
Closes #7

### DIFF
--- a/.github/actions/archive-paparazzi-reports/action.yml
+++ b/.github/actions/archive-paparazzi-reports/action.yml
@@ -1,0 +1,16 @@
+name: '📦 Archive Paparazzi reports'
+description: 'Archive Paparazzi reports and deltas'
+runs:
+  using: composite
+  steps:
+    - name: 'Archive report'
+      uses: actions/upload-artifact@v3
+      with:
+        name: paparazzi-html
+        path: '**/build/reports/paparazzi'
+    - name: 'Archive deltas'
+      if: ${{ hashFiles('**/out/failures/**.png') != '' }}
+      uses: actions/upload-artifact@v3
+      with:
+        name: paparazzi-delta
+        path: '**/out/failures/**.png'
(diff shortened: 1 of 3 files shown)


.github/actions/archive-paparazzi-reports/action.yml
```diff
@@ -0,0 +1,16 @@
...
+      uses: actions/upload-artifact@v3
+      with:
+        name: paparazzi-html
+        path: '**/build/reports/paparazzi'
+    - name: 'Archive deltas'
+      if: ${{ hashFiles('**/out/failures/**.png') != '' }}
+      uses: actions/upload-artifact@v3
+      with:
+        name: paparazzi-delta
+        path: '**/out/failures/**.png'
```
soulcramer, line 16: SHouldn't this be '**/out/failures/delta-**.png' or do we want to upload all screenshots?

SimonMarquis: I suppose we want both. `Delta-*` are composited images that could differ from the real content.

soulcramer: Actually they have the previous one on the left side | the delta in the middle | the new one on the right side so imo we can just use these

SimonMarquis: Yes I know they combine both images, but the resulting composition does not correspond exactly to what as been produced by paparazzi. So I prefer we keep both files for now.

soulcramer: Good to me

SimonMarquis:
I think this looks good enough for now, cc @soulcramer

soulcramer (approved):
LGTM :awesome:

Issue #117, opened 2023-02-20
Move SparkIcons to this repo


soulcramer:
Since we use this in some components we need it before migrating them

SimonMarquis:
On this subject, we currently have

```kotlin
abstract class SparkIcon(@DrawableRes override val drawableId: Int) : IconDrawableRes
```

Is it abstract by design? Do we allow custom SparkIcons implementations to exist outside of this module?
Because at the moment, the autocompletion show this:

![image](https://user-images.githubusercontent.com/1921278/220076447-a9031e71-d2a4-4cd3-b0c6-fb8a9d5785cf.png)

If not, we probably should make it a sealed class (and the subclasses as well.

```kotlin
sealed class SparkIcon constructor(@DrawableRes override val drawableId: Int) : IconDrawableRes
```

This would make it simpler to use IMO:
![image](https://user-images.githubusercontent.com/1921278/220076797-47a273ec-3894-4895-bfdb-33586ddc5eb5.png)

soulcramer:
It makes perfect sense!

I used an abstract class initially like you said to allow verticals to extend `SparkIcon` to use their own Icons on components that require a `SparkIcon` to ease Compose migration but now that we have a real Design System initiative I think we can enforce these icons until the verticals have a "real" need for customization

soulcramer:
I've made a first PR #126 where we can put all of this in it but do we want multi-module support before merging it? wdyt @SimonMarquis ?

SimonMarquis:
You can copy paste the configuration for now. I'm in the process of creating the convention plugins at the moment :)

PR #126, opened 2023-02-22
chore: Create empty spark-icons module

soulcramer:
## 🧑‍ Changes description

Add a new :spark-icons module  and make :spark depend from it

## 🤔 Context

It will contains our SparkIcon class with the drawables used by the design system

## 👋 Other info

Initialization of #117


gradle/libs.versions.toml
```diff
@@ -30,7 +31,9 @@ airbnb-showkase = { module = "com.airbnb.android:showkase", version.ref = "showk
 airbnb-showkase-processor = { module = "com.airbnb.android:showkase-processor", version.ref = "showkase" }
 
 android-documentation-plugin = { module = "org.jetbrains.dokka:android-documentation-plugin", version.ref = "androidDocumentationPlugin" }
+# Artifact: https://mvnrepository.com/artifact/androidx.appcompat/appcompat
```
SimonMarquis, line 34: IMO this info is useless. We should either redirect to a public github project or changelog. Also, IntelliJ now directly shows the available updates.

spark-icons/build.gradle.kts
```diff
@@ -0,0 +1,79 @@
...
+    lint {
+        warningsAsErrors = true
+        sarifReport = true
+    }
+}
+
+dependencies {
+    implementation(platform(libs.androidx.compose.bom))
+    implementation(libs.androidx.compose.ui)
+    api(libs.androidx.appCompat.ressources) // Needed for compat vector drawables
```
SimonMarquis, line 78: Does it needs to be exposed as an api though?

spark-icons/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
```
SimonMarquis, line 1: Could be removed I think

soulcramer: Done in c995245

spark-icons/build.gradle.kts
soulcramer: I'm not sure but we can change it back to api later if needed 👍 
c995245

gradle/libs.versions.toml
soulcramer: I personally use androidx.tech to get the changelog & the sources is it good enough or should we reference the androidx Github?
c995245

SimonMarquis (approved):

PR #142, opened 2023-02-24
feat: Add SparkIcon type

### LINKED ISSUE #117: Move SparkIcons to this repo

soulcramer:
## 🧑‍ Changes description

Add SparkIcon type in :spark-icons module 

## 🤔 Context

This will make it more easy to use the same icon that the Design use and enforce these icons to be use with Spark components

## 👋 Other info

Close #117 
Followup to #126


spark-icons/src/main/kotlin/com/adevinta/spark/icons/SparkIcon.kt
```diff
@@ -0,0 +1,637 @@
...
+        public sealed class Lock(@DrawableRes override val drawableId: Int) : Infos(drawableId) {
+            public object Default : Lock(R.drawable.spark_icons_lock)
+            public object Outline : Lock(R.drawable.spark_icons_lock_outline)
+            public object Open : Lock(R.drawable.spark_icons_open)
+        }
+
+        public object Warning : Infos(R.drawable.spark_icons_warning)
+    }
+
+    public abstract class Delivery(@DrawableRes override val drawableId: Int) : SparkIcon(drawableId) {
```
SimonMarquis, line 468:
```suggestion
    public class Delivery(@DrawableRes override val drawableId: Int) : SparkIcon(drawableId) {
```

spark-icons/src/main/kotlin/com/adevinta/spark/icons/SparkIcon.kt
```diff
@@ -0,0 +1,637 @@
...
+        //        public object DeliveryFast : Delivery(R.drawable.spark_icons_warning)
```
SimonMarquis, line 469: What is this icon? We should remove it if we don't have it IMO.

spark-icons/src/main/kotlin/com/adevinta/spark/icons/SparkIcon.kt
```diff
@@ -0,0 +1,637 @@
...
+            }
+
+            public sealed class Open(@DrawableRes override val drawableId: Int) : Mailbox(drawableId) {
+                public object Down : Open(R.drawable.spark_icons_delivery_mailbox_open_down)
+                public object Up : Open(R.drawable.spark_icons_delivery_mailbox_open_up)
+            }
+        }
+
+        public object LivraisonPerso : Delivery(R.drawable.spark_icons_delivery_sparkles)
+        public abstract class Partner(@DrawableRes override val drawableId: Int) : SparkIcon(drawableId)
```
SimonMarquis, line 486:
```suggestion
        public class Partner(@DrawableRes override val drawableId: Int) : SparkIcon(drawableId)
```

spark-icons/src/main/kotlin/com/adevinta/spark/icons/SparkIcon.kt
```diff
@@ -0,0 +1,637 @@
...
+    public sealed class Value(@DrawableRes override val drawableId: Int) : SparkIcon(drawableId) {
+        public object Minus : Value(R.drawable.spark_icons_minus_outline)
+
+        public sealed class More(@DrawableRes override val drawableId: Int) : Value(drawableId) {
+            public object Default : More(R.drawable.spark_icons_more)
+            public object Outline : More(R.drawable.spark_icons_more_outline)
+        }
+    }
+
+    public abstract class Logo(@DrawableRes override val drawableId: Int) : SparkIcon(drawableId)
```
SimonMarquis, line 636: IMO This should not be call Logo, but something more generic. For instance `Custom`. WDYT?

```suggestion
    public abstract class Custom(@DrawableRes override val drawableId: Int) : SparkIcon(drawableId)
```

SimonMarquis (dismissed):

SimonMarquis (approved):

PR #147, opened 2023-02-27
feat: Add support for WindowSizeClass in Layout tokens

soulcramer:
## 🧑‍ Changes description

Add `windowSize` property on `Layout` tokens

## 🤔 Context

This will help us build component that can react to different form factors

## 📸 Screenshots

For exemple we used it internally for now to build an adaptative full screen modal:
image
image
image
image


gradle/libs.versions.toml
```diff
@@ -37,6 +38,7 @@ android-documentation-plugin = { module = "org.jetbrains.dokka:android-documenta
 androidx-activity-compose = { group = "androidx.activity", name = "activity-compose", version.ref = "androidx-activity" }
 # https://androidx.tech/artifacts/appcompat/appcompat-resources/
 androidx-appCompat-resources = { module = "androidx.appcompat:appcompat-resources", version.ref = "androidx-appCompat" }
+androidx-core = { module = "androidx.core:core-ktx", version.ref = "androidx-core" }
```
SimonMarquis, line 41: It seems we don't need `core-ktx` since only `androidx.window:window` appears in the runtime classpath. Can we change this dependency?

soulcramer: Yep I will change it 👍

SimonMarquis (approved):

PR #162, opened 2023-02-28
chore: Add Switch component

soulcramer:
## 🧑‍ Changes description

Add Switch component

## 🤔 Context

We couldn't integrate it at first since it use the Icon component that was not present. This is not the cas anymore so here we are now

## 📸 Screenshots

They're added as files in this review


spark/src/main/kotlin/com/adevinta/spark/components/toggles/Switch.kt
```diff
@@ -0,0 +1,190 @@
...
+}
+
+/**
+ *
+ * Switches are the preferred way to adjust settings. They're used to control binary options – think On/Off or True/False.
+ *
+ *  - Toggle a single item on or off.
+ *  - Immediately activate or deactivate something.
+ *
+ * @see [SparkSwitch] if you require color customization between states. Be aware that this is still an internal composable so if you need such state contact the brikke team
```
SimonMarquis, line 64:
```suggestion
 * @see [SparkSwitch] if you require color customization between states. Be aware that this is still an internal composable so if you need such state contact the Spark team
```

spark/src/main/kotlin/com/adevinta/spark/components/toggles/Switch.kt
```diff
@@ -0,0 +1,190 @@
...
```
SimonMarquis, line 99:
```suggestion
 * @see [SparkSwitch] if you require color customization between states. Be aware that this is still an internal composable so if you need such state contact the Spark team
```

spark/src/main/kotlin/com/adevinta/spark/components/toggles/Switch.kt
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * @param checked whether or not this component is checked
+ * @param onCheckedChange callback to be invoked when Switch is being clicked, therefore the change of checked state is requested. If null, then this is passive and relies entirely on a higher-level component to control the "checked" state.
+ * @param modifier Modifier to be applied to the layout of the switch layout
+ * @param enabled whether the component is enabled or grayed out
+ * @param interactionSource the [MutableInteractionSource] representing the stream of
+ * [Interaction]s for this Switch. You can create and pass in your own remembered
+ * [MutableInteractionSource] if you want to observe [Interaction]s and customize the
+ * appearance / behavior of this Switch in different [Interaction]s.
+ * @param startContent The start content displayed before the checkbox, Usually a Text composable
```
SimonMarquis, line 109:
```suggestion
 * @param startContent The start content displayed before the switch, Usually a Text composable
```

SimonMarquis (approved):

SimonMarquis (approved):

PR #217, opened 2023-03-02
feat: Add AnnotatedString support

### LINKED ISSUE #204: Add the ability to annotate a string with primary/secondary/etc.
We already have the `val annotatedString = annotatedStringResource(R.string.test)` function to generate an annotated string based on the html tags in a text ressource but currently it doesn't handle the color changes with the spark tokens

soulcramer:
## 🧑‍ Changes description

Add stylized formatted and localized string ressource support to compose with `AnnotatedString`.
This can also handle the DS color & typography tokens

## 🤔 Context

For exemple if we have the following strings:

- **en**:
``` xml
<string name="spark_annotatedStringResource_test"><b>Best</b> <annotation typography="large">practices</annotation> for text on <annotation color="primary">Android</annotation></string>
```

- **fr**:
``` xml
<string name="spark_annotatedStringResource_test">Les <b>Meilleures</b> <annotation typography="large">pratiques</annotation> pour les textes sur <annotation color="primary">Android</annotation></string>
```

it will display like this:

 - **en**:
 
image

- **fr**:

image

close #204


spark/src/main/kotlin/com/adevinta/spark/res/SparkStringAnnotations.kt
```diff
@@ -0,0 +1,80 @@
...
+            "bodyImportant" -> token.bodyImportant.toSpanStyle()
+            "body" -> token.body.toSpanStyle()
+            "smallImportant" -> token.smallImportant.toSpanStyle()
+            "small" -> token.small.toSpanStyle()
+            "extraSmallImportant" -> token.extraSmallImportant.toSpanStyle()
+            "extraSmall" -> token.extraSmall.toSpanStyle()
+            "button" -> token.button.toSpanStyle()
+            else -> null
+        }
+    }
```
SimonMarquis, line 64: DRY

```suggestion
        public override fun toSpanStyle(value: String, token: SparkTypography): SpanStyle? = when (value) {
            "title1" -> token.title1
            "title2" -> token.title2
            "title3" -> token.title3
            "bodyImportant" -> token.bodyImportant
            "body" -> token.body
            "smallImportant" -> token.smallImportant
            "small" -> token.small
            "extraSmallImportant" -> token.extraSmallImportant
            "extraSmall" -> token.extraSmall
            "button" -> token.button
            else -> null
        }?.toSpanStyle()
    }
```

spark/src/main/kotlin/com/adevinta/spark/res/SparkStringAnnotations.kt
```diff
@@ -0,0 +1,80 @@
...
+            "alert" -> SpanStyle(color = token.alert)
+            "error" -> SpanStyle(color = token.error)
+            "info" -> SpanStyle(color = token.info)
+            "neutral" -> SpanStyle(color = token.neutral)
+            else -> null
+        }
+    }
+
+    public object Typography : SparkStringAnnotations<SparkTypography> {
+        override fun toString(): String = "typography"
```
SimonMarquis, line 49: nit: I would have done the same mapping for this parent layer.

spark/src/main/kotlin/com/adevinta/spark/res/SparkStringAnnotations.kt
```diff
@@ -0,0 +1,80 @@
...
+            "primary" -> SpanStyle(color = token.primary)
+            "secondary" -> SpanStyle(color = token.secondary)
+            "success" -> SpanStyle(color = token.success)
+            "alert" -> SpanStyle(color = token.alert)
+            "error" -> SpanStyle(color = token.error)
+            "info" -> SpanStyle(color = token.info)
+            "neutral" -> SpanStyle(color = token.neutral)
+            else -> null
+        }
+    }
```
SimonMarquis, line 46: DRY
```suggestion
        public override fun toSpanStyle(value: String, token: SparkColors): SpanStyle? = when (value) {
            "primary" -> token.primary
            "secondary" -> token.secondary
            "success" -> token.success
            "alert" -> token.alert
            "error" -> token.error
            "info" -> token.info
            "neutral" -> token.neutral
            else -> null
        }
    }?.let(::SpanStyle)
```

spark/src/main/res/values/strings.xml
```diff
@@ -30,4 +30,7 @@
     <string name="spark_rating_a11y">Note de %1$.1f</string>
 
     <string name="spark_user_avatar_content_description">Photo de profil</string>
+
+    <string name="spark_annotatedStringResource_test"><b>Best</b> <annotation typography="large">practices</annotation> for text on <annotation color="primary">Android</annotation></string>
```
SimonMarquis, line 34: nit: this should go into a test source. We don't want to ship this in the aar.

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+@ReadOnlyComposable
+public fun resources(): Resources {
+    LocalConfiguration.current
+    return LocalContext.current.resources
+}
+
+public fun Spanned.toHtmlWithoutParagraphs(): String {
+    return HtmlCompat.toHtml(this, HtmlCompat.TO_HTML_PARAGRAPH_LINES_CONSECUTIVE)
+        .substringAfter("<p dir=\"ltr\">").substringBeforeLast("</p>")
+}
```
SimonMarquis, line 60: What is this trying to achieve? This is almost certain to fail in unexpected ways.

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+import com.adevinta.spark.SparkTheme
+import com.adevinta.spark.tokens.SparkColors
+import com.adevinta.spark.tokens.SparkTypography
+
+
+// FIXME: There is no official way to do this yet so we're waiting for
+//   https://issuetracker.google.com/issues/139320238 to be fixed
+@Composable
+@ReadOnlyComposable
+public fun resources(): Resources {
```
SimonMarquis, line 52: IMO we should keep this internal for now

```suggestion
internal fun resources(): Resources {
```

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+    LocalConfiguration.current
+    return LocalContext.current.resources
+}
+
+public fun Spanned.toHtmlWithoutParagraphs(): String {
+    return HtmlCompat.toHtml(this, HtmlCompat.TO_HTML_PARAGRAPH_LINES_CONSECUTIVE)
+        .substringAfter("<p dir=\"ltr\">").substringBeforeLast("</p>")
+}
+
+public fun Resources.getText(@StringRes id: Int, vararg args: Any): CharSequence {
```
SimonMarquis, line 62: This is unexpected, Spanned are not part of what is supported by Android's Resource.getText, and this might introduce many errors from the caller's perspective.

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+    val escapedArgs = args.map {
+        if (it is Spanned) it.toHtmlWithoutParagraphs() else it
+    }.toTypedArray()
+    val resource = SpannedString(getText(id))
```
SimonMarquis, line 66: this variable should be renamed, as this is not a resource.

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+    val htmlResource = resource.toHtmlWithoutParagraphs()
+    val formattedHtml = String.format(htmlResource, *escapedArgs)
+    return HtmlCompat.fromHtml(formattedHtml, HtmlCompat.FROM_HTML_MODE_LEGACY)
+}
+
+@Composable
+public fun annotatedStringResource(@StringRes id: Int, vararg formatArgs: Any): AnnotatedString {
```
SimonMarquis, line 73: The naming of these functions do not represent their return type. They don't return a resource but an annotated string, this feels weird to me.

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+
+    val colors = SparkTheme.colors
+    val typography = SparkTheme.typography
+    return remember(id) {
+        val text = resources.getText(id)
+        spannableStringToAnnotatedString(text, density, colors, typography)
+    }
+}
+
+private fun spannableStringToAnnotatedString(
```
SimonMarquis, line 98: This would be a perfect fit for an extension name `.toAnnotatedString()`

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+    text: CharSequence,
+    density: Density,
+    colors: SparkColors,
+    typography: SparkTypography,
+): AnnotatedString {
+    return if (text is Spanned) {
```
SimonMarquis, line 104: Here, do an early exit to reduce the amount of indentation

```
if (test !is Spanned) return AnnotatedString(text.toString())

...

return ...
```

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+        with(density) {
```
SimonMarquis, line 105: IMO we should not create more indentation here to the minimal benefit of having density as a receiver type.

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+@Preview(
+    locale = "fr-rFR",
+)
+@Composable
+private fun AnnotatedStringResourcePreview() {
+    PreviewTheme {
+        val annotatedString = annotatedStringResource(R.string.spark_annotatedStringResource_test)
+        Text(
+            annotatedString,
+        )
```
SimonMarquis, line 236:
```suggestion
        Text(annotatedString)
```

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+private fun spannableStringToAnnotatedString(
+    text: CharSequence,
+    density: Density,
+    colors: SparkColors,
+    typography: SparkTypography,
+): AnnotatedString {
+    return if (text is Spanned) {
+        with(density) {
+            buildAnnotatedString {
+                append((text.toString()))
```
SimonMarquis, line 107:
```suggestion
                append(text.toString())
```

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,238 @@
...
+                text.getSpans(0, text.length, Any::class.java).forEach {
```
SimonMarquis, line 108: Nit: extract this inner loop into a `AnnotatedString.Builder` extension to reduce indentation and complexity.

SimonMarquis: For example:


```kotlin

private fun spannableStringToAnnotatedString(
    text: CharSequence,
    density: Density,
    colors: SparkColors,
    typography: SparkTypography,
): AnnotatedString {
    if (text !is Spanned) return AnnotatedString(text.toString())
    return buildAnnotatedString {
        append(text.toString())
        text.getSpans(0, text.length, Any::class.java).forEach {
            buildWithSpan(it, text.getSpanStart(it), text.getSpanEnd(it))
        }
    }
}

private fun AnnotatedString.Builder.buildWithSpan(it: Any, start: Int, end: Int) = when (it) {
    is StyleSpan -> buildWithStyleSpan(it, start, end)
    is ...
}


private fun AnnotatedString.Builder.buildWithStyleSpan(it: StyleSpan, start: Int, end: Int) = when (it.style) {
    NORMAL -> FontWeight.Normal to FontStyle.Normal
    BOLD -> FontWeight.Bold to FontStyle.Normal
    ITALIC -> FontWeight.Normal to FontStyle.Italic
    BOLD_ITALIC -> FontWeight.Bold to FontStyle.Italic
    else -> null
}?.let { (fontWeight, fontStyle) ->
    addStyle(SpanStyle(fontWeight = fontWeight, fontStyle = fontStyle), start, end)
}
```

spark/src/main/kotlin/com/adevinta/spark/res/SparkStringAnnotations.kt
```diff
@@ -0,0 +1,91 @@
...
+        private fun String.asSparkAnnotation(): SparkStringAnnotations<*>? {
+            return when (this) {
+                "color" -> Colors
+                "typography" -> Typography
+                else -> null.also { _ ->
+                    Log.d("StringResources", "Annotation  $this is not supported by spark")
+                }
+            }
+        }
+    }
```
SimonMarquis, line 90: IMO we don't need this extra layer of API as it does not bring any benefit vs directly calling

```kotlin
        public fun toSpanStyle(
            key: String,
            value: String,
            colors: SparkColors,
            typography: SparkTypography,
        ): SpanStyle {
            return when (key) {
                "color" -> Colors.toSpanStyle(value, colors)
                "typography" -> Typography.toSpanStyle(value, typography)
                else -> /* log */ SpanStyle()
            }
        }
```

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
soulcramer: Lets make it internal

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
soulcramer: Actually we're following the naming for ressources from Compose like these api:
``` kotlin
stringResource(id = )
stringArrayResource(id = )
pluralStringResource(id = , count = )
imageResource(id= )
```

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,201 @@
...
+
+@Composable
+@ReadOnlyComposable
+internal fun resources(): Resources {
+    LocalConfiguration.current
+    return LocalContext.current.resources
+}
+
+/**
+ * Converts a [Spanned] to a [String] without the surrounding `<p dir=”rtl | ltr” style=”…”>` tags as we don't support
```
SimonMarquis, line 98: wrong chars
```suggestion
 * Converts a [Spanned] to a [String] without the surrounding `<p dir="rtl | ltr" style="…">` tags as we don't support
```

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,201 @@
...
+
+        is RelativeSizeSpan -> SpanStyle(fontSize = it.sizeChange.em)
+        is StrikethroughSpan -> SpanStyle(textDecoration = TextDecoration.LineThrough)
+        is UnderlineSpan -> SpanStyle(textDecoration = TextDecoration.Underline)
+        is SuperscriptSpan -> SpanStyle(baselineShift = BaselineShift.Superscript)
+        is SubscriptSpan -> SpanStyle(baselineShift = BaselineShift.Subscript)
+        is ForegroundColorSpan -> SpanStyle(color = Color(it.foregroundColor))
+        is Annotation -> {
+            SparkStringAnnotations.toSpanStyle(it.key, it.value, colors, typography)
+        }
```
SimonMarquis, line 162:
```suggestion
        is Annotation -> SparkStringAnnotations.toSpanStyle(it.key, it.value, colors, typography)
```

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,199 @@
...
+
+        is RelativeSizeSpan -> SpanStyle(fontSize = it.sizeChange.em)
+        is StrikethroughSpan -> SpanStyle(textDecoration = TextDecoration.LineThrough)
+        is UnderlineSpan -> SpanStyle(textDecoration = TextDecoration.Underline)
+        is SuperscriptSpan -> SpanStyle(baselineShift = BaselineShift.Superscript)
+        is SubscriptSpan -> SpanStyle(baselineShift = BaselineShift.Subscript)
+        is ForegroundColorSpan -> SpanStyle(color = Color(it.foregroundColor))
+        is Annotation -> SparkStringAnnotations.toSpanStyle(it.key, it.value, colors, typography)
+
+        else -> SpanStyle()
```
SimonMarquis, line 162: Do we want to create empty Spans? Shouldn't we return instead?

spark/src/main/kotlin/com/adevinta/spark/res/SparkStringAnnotations.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.adevinta.spark.tokens.SparkColors
+import com.adevinta.spark.tokens.SparkTypography
+
+/**
+ * Represents a set of annotations supported by spark that can be used in a string resource.
+ */
+public interface SparkStringAnnotations<T : Any> {
+
+    public fun toSpanStyle(value: String, token: T): SpanStyle?
+
```
SimonMarquis, line 36: I'm still not convinced this brings any benefit.
We never use the type SparkStringAnnotations, and subclasses don't hold any value.
IMO this would be way simpler.

```
        public fun Annotation.toSpanStyle(
            key: String,
            value: String,
            colors: SparkColors,
            typography: SparkTypography,
        ): SpanStyle {
            return when (key) {
                "color" -> value.toColorSpanStyle(colors)
                "typography" -> value.toTypographySpanStyle(typography)
                else -> null.also { _ ->
                    Log.d("StringResources", "Annotation $this is not supported by spark")
                }
            }
        }
```

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,200 @@
...
+    return when (it.style) {
+        Typeface.NORMAL -> FontWeight.Normal to FontStyle.Normal
+        Typeface.BOLD -> FontWeight.Bold to FontStyle.Normal
+        Typeface.ITALIC -> FontWeight.Normal to FontStyle.Italic
+        Typeface.BOLD_ITALIC -> FontWeight.Bold to FontStyle.Italic
+        else -> null
+    }?.let { (fontWeight, fontStyle) ->
+        SpanStyle(fontWeight = fontWeight, fontStyle = fontStyle)
+    } ?: SpanStyle()
+}
```
SimonMarquis, line 178: To keep the same naming, should we also update like this? (and remove the empty span)

```suggestion
private fun StyleSpan.to SpanStyle(it: StyleSpan): SpanStyle? {
    return when (it.style) {
        Typeface.NORMAL -> FontWeight.Normal to FontStyle.Normal
        Typeface.BOLD -> FontWeight.Bold to FontStyle.Normal
        Typeface.ITALIC -> FontWeight.Normal to FontStyle.Italic
        Typeface.BOLD_ITALIC -> FontWeight.Bold to FontStyle.Italic
        else -> null
    }?.let { (fontWeight, fontStyle) ->
        SpanStyle(fontWeight = fontWeight, fontStyle = fontStyle)
    }
}
```

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,200 @@
...
+
+private fun buildWithTypefaceSpan(it: TypefaceSpan): SpanStyle = SpanStyle(
+    fontFamily = when (it.family) {
+        FontFamily.SansSerif.name -> FontFamily.SansSerif
+        FontFamily.Serif.name -> FontFamily.Serif
+        FontFamily.Monospace.name -> FontFamily.Monospace
+        FontFamily.Cursive.name -> FontFamily.Cursive
+        else -> FontFamily.Default
+    },
+)
```
SimonMarquis, line 188: Same here:

```suggestion
private fun TypefaceSpan.toSpanStyle(): SpanStyle = SpanStyle(
    fontFamily = when (it.family) {
        FontFamily.SansSerif.name -> FontFamily.SansSerif
        FontFamily.Serif.name -> FontFamily.Serif
        FontFamily.Monospace.name -> FontFamily.Monospace
        FontFamily.Cursive.name -> FontFamily.Cursive
        else -> FontFamily.Default
    },
)
```

spark/src/main/kotlin/com/adevinta/spark/res/SparkStringAnnotations.kt
```diff
@@ -0,0 +1,89 @@
...
+    public fun toSpanStyle(
+        key: String,
+        value: String,
+        colors: SparkColors,
+        typography: SparkTypography,
+    ): SpanStyle {
+        return when (key) {
+            "color" -> value.toColorSpanStyle(colors)
+            "typography" -> value.toTypographySpanStyle(typography)
+            else -> SpanStyle().also { _ ->
```
SimonMarquis, line 47: This should return a null value

spark/src/main/kotlin/com/adevinta/spark/res/SparkStringAnnotations.kt
```diff
@@ -0,0 +1,89 @@
...
+    private fun String.toColorSpanStyle(token: SparkColors): SpanStyle = when (this) {
+        "primary" -> token.primary
+        "secondary" -> token.secondary
+        "success" -> token.success
+        "alert" -> token.alert
+        "error" -> token.error
+        "info" -> token.info
+        "neutral" -> token.neutral
+        else -> null
+    }?.let(::SpanStyle) ?: SpanStyle().also { _ ->
```
SimonMarquis, line 65: null here as well

spark/src/main/kotlin/com/adevinta/spark/res/SparkStringAnnotations.kt
```diff
@@ -0,0 +1,89 @@
...
+        "largeImportant" -> token.largeImportant
+        "bodyImportant" -> token.bodyImportant
+        "body" -> token.body
+        "smallImportant" -> token.smallImportant
+        "small" -> token.small
+        "extraSmallImportant" -> token.extraSmallImportant
+        "extraSmall" -> token.extraSmall
+        "button" -> token.button
+        else -> null
+    }?.toSpanStyle() ?: SpanStyle().also { _ ->
```
SimonMarquis, line 86: same

spark/src/main/res/values/strings.xml
soulcramer: I'll remove it once we can render all these styles in the sample app as the layout lib doesn't support rendering of annotated string yet

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,198 @@
...
+            SpanStyle(fontSize = if (it.dip) it.size.dp.toSp() else it.size.toSp())
+        }
+
+        is RelativeSizeSpan -> SpanStyle(fontSize = it.sizeChange.em)
+        is StrikethroughSpan -> SpanStyle(textDecoration = TextDecoration.LineThrough)
+        is UnderlineSpan -> SpanStyle(textDecoration = TextDecoration.Underline)
+        is SuperscriptSpan -> SpanStyle(baselineShift = BaselineShift.Superscript)
+        is SubscriptSpan -> SpanStyle(baselineShift = BaselineShift.Subscript)
+        is ForegroundColorSpan -> SpanStyle(color = Color(it.foregroundColor))
+        is Annotation -> SparkStringAnnotations.toSpanStyle(it.key, it.value, colors, typography) ?: return
```
SimonMarquis, line 161: We should keep the `Annotation` type instead of extracting key/value.

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -0,0 +1,198 @@
...
+
+private fun StyleSpan.toSpanStyle(): SpanStyle = when (style) {
+    Typeface.NORMAL -> FontWeight.Normal to FontStyle.Normal
+    Typeface.BOLD -> FontWeight.Bold to FontStyle.Normal
+    Typeface.ITALIC -> FontWeight.Normal to FontStyle.Italic
+    Typeface.BOLD_ITALIC -> FontWeight.Bold to FontStyle.Italic
+    else -> null
+}?.let { (fontWeight, fontStyle) ->
+    SpanStyle(fontWeight = fontWeight, fontStyle = fontStyle)
+} ?: SpanStyle()
```
SimonMarquis, line 176: This should not create an empty `SpanStyle`.

SimonMarquis (approved):

spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
SimonMarquis: Can we explain and test this method?

SimonMarquis (approved):

PR #306, opened 2023-03-31
fix: Modify `Image` component color tokens

soulcramer:
## 📋 Changes description

Modify `Image` component color tokens to use neutral Container instead of neutral as background

## 🤔 Context

The rendering of the states was too different from the one we previously had, turns out we're using the wrong token for background.

## 📸 Screenshots

Present in the files


SimonMarquis (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ## [Unreleased]
 
+* Modify the `Image` fallback states backgroudn and icon colors #306
```
SimonMarquis, line 7:
```suggestion
* Modify the `Image` fallback states background and icon colors #306
```

SimonMarquis (approved):

PR #310, opened 2023-04-04
feat: Add Legacy option to use previous DS style

soulcramer:
## 📋 Changes description

Add new param to `SparkTheme` & new CompositionLocal `LocalLegacyStyle`

## 🤔 Context

Since we're migrating screens with Spark on the Adevinta Platform we don't want a screen to have a component with different UI if a part of the screen is not yet migrated.

This 'll allow us to have component with a UI close enough from the previous one to avoid huge difference between components .

### DIFF
--- a/spark/src/main/kotlin/com/adevinta/spark/SparkTheme.kt
+++ b/spark/src/main/kotlin/com/adevinta/spark/SparkTheme.kt
@@ -238,7 +238,7 @@ internal val LocalHighlightToken = staticCompositionLocalOf { false }
 internal val LocalHighlightComponents = staticCompositionLocalOf { false }
 
 /**
- * CompositionLocal that makes the components use the legacy style from the previous DS to make it easier for the Adevinta Platform teams.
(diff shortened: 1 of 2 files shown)


spark/src/main/kotlin/com/adevinta/spark/SparkTheme.kt
```diff
@@ -232,3 +236,9 @@ internal val LocalHighlightToken = staticCompositionLocalOf { false }
  * Setting it to true show an overlay on spark components.
  */
 internal val LocalHighlightComponents = staticCompositionLocalOf { false }
+
+/**
+ * CompositionLocal that makes the components use the legacy style from the previous DS to make it easier for the Adevinta Platform teams.
```
MarinaRomanova, line 241:
```suggestion
 * CompositionLocal that makes the components use the legacy style from the previous DS to make it easier for the Adevinta Platform teams
```

MarinaRomanova (approved):

PR #321, opened 2023-04-12
fix: Fix the loading flickering on Image component

### LINKED ISSUE #316: [Bug]: Skeleton visible on HorizontalPager with Spark Image

1.0.0

### 🐛 What happened?

When using HorizontalPager with the Spark image component, we can see a quick grey background on the image.

Video:

https://user-images.githubusercontent.com/4458225/230617306-1ab4bc75-83fb-4fa2-ae83-083c572811e9.mov

### Relevant logcat output

_No response_

### Code of Conduct

soulcramer:
## 📋 Changes description

Add a 200ms delay before showing the content or the loading and if in loading state then we wait 500ms before hiding it to avoid the quick flickering between the loading state and the other ones.

## ✅ Checklist

- [x] Link to GitHub issues it solves. close #316 

## 🗒️ Other info

We might want to open the delays in the api 🤔 ? Or we also can make this logic into a composable if we want to reuse it for other components


spark/src/main/kotlin/com/adevinta/spark/components/image/Image.kt
```diff
@@ -148,6 +149,8 @@ internal fun SparkImage(
  * @param filterQuality Sampling algorithm applied to the image when it is scaled and drawn
  * into the destination. The default is [FilterQuality.Low] which scales using a bilinear
  * sampling algorithm
+ * @param withPlaceholder Don't show the shimmering placeholder in cases where the images is in fullscreen which
+ * would make a big flash when the image is loaded too quick (~less than 500ms)
```
SimonMarquis, line 153: This documentation should be re-worded with the opposite meaning.
`withPlaceholder` default value is true, and requesting it would `show the shimmering placeholder`

SimonMarquis (changes requested):

SimonMarquis (approved):

SimonMarquis (approved):

PR #327, opened 2023-04-13
feat: Add fullscreen modal component as experimental

### LINKED ISSUE #205: Add a full screen dialog component
This is almost done internaly but we still need to figure out the Dialog size in a phone form factor.

We can have a sneak peek in #147

soulcramer:
## 📋 Changes description

Add fullscreen modal component as experimental and change the api of the conditions modifiers 

## 🤔 Context

This type of dialog/modal is used a lot in Polaris so we provide a generic way to show them and handle the different form factors layout for them.

## 📸 Screenshots

image

image

image

image

close #205


spark/src/main/kotlin/com/adevinta/spark/components/dialog/ModalFullScreenScaffold.kt
```diff
@@ -0,0 +1,362 @@
...
+    Scaffold(
+        modifier = modifier.fillMaxSize(),
+        snackbarHost = snackbarHost,
+        topBar = {
+            TopAppBar(
+                navigationIcon = {
+                    IconButton(onClick = onClose) {
+                        Icon(
+                            sparkIcon = SparkIcon.Navs.Close,
+                            contentDescription = null, // FIXME-scott.rayapoulle.ext (13-36-2023): Add description
```
MarinaRomanova, line 270: either we should fix it or expose a param

MarinaRomanova (changes requested):

PR #356, opened 2023-04-25
feat: Adapt Switch Toggle to respect Spark specs (#302)

MarinaRomanova:
## 📋 Changes description
- Update Switch Toggle component to respect Spark specs. 
- Update SparkToggleLabelledContainer to respect the padding between components as per Spark specs.

## 📸 Screenshots
Generated by paparazzi in this PR files

## 🗒️ Other info
Like noted in the documentation, we're missing the layout part and the form state part as their are not priorities for now but we intend to add them when required.

close[#302]


shkschneider (approved):
LGTM

PR #369, opened 2023-05-04
refactor: Update Switch Toggle component

### LINKED ISSUE #370: [Switch] Update Default state and Icon customisation
The decision has been made to not include icons in the default state and to not allow a state where only one icon is present.

MarinaRomanova:
## 📋 Changes description
- Make the switch with no icons a default version
- Use a data class to wrap checked/unchecked icons

close #370


spark/src/main/kotlin/com/adevinta/spark/components/toggles/Switch.kt
```diff
@@ -178,6 +173,14 @@ public fun SwitchLabelled(
         content = content,
     )
 }
+/**
+ * @property checkedIcon icon to be used for the thumb in checked state
+ * @property uncheckedIcon icon to be used for the thumb in unchecked state
+ */
+public data class IconState(
```
soulcramer, line 180: What about `SwitchIcons` as a name

spark/src/main/kotlin/com/adevinta/spark/components/toggles/Switch.kt
```diff
@@ -178,6 +173,14 @@ public fun SwitchLabelled(
...
+    val checkedIcon: SparkIcon = SparkIcon.Toggles.Check.Simple,
```
soulcramer, line 181: since we already have icon as the class name imo we could use only `checked` and `unchecked`

spark/src/main/kotlin/com/adevinta/spark/components/toggles/Switch.kt
```diff
@@ -189,46 +192,42 @@ internal fun AllStatesSwitchPreview(
 ) {
     val (theme, userType) = param
     PreviewTheme(theme, userType) {
+        Row {
+            Switch(checked = true, onCheckedChange = {}, enabled = true, iconState = IconState())
```
soulcramer, line 196: We can also add it to `SwitchDefaults.icons`

soulcramer (approved):

PR #379, opened 2023-05-16
chore: Improve Chips.md

MarinaRomanova:
## 📋 Changes description
Improve Chips documentation by providing details how to handle selection for accessibility services.
Delete repetitions.


spark/src/main/kotlin/com/adevinta/spark/components/chips/Chip.md
```diff
@@ -213,4 +205,12 @@ Chip(
     text = "Chip",
     leadingIcon = if (selected) SparkIcon.Toggles.Check.Simple else null
 )
+```
+
+To correctly handle selectable state, pass the following modifier to components 
+ ```kotlin
+fun Modifier.semantics {
```
soulcramer, line 212: I think the `selectable` or the `togglable` modifer provide this already, did you looked into it? If not we might want to verify this

MarinaRomanova: Yes, there is a `clickable` part in it + `semantics`. But we have already onClick param, so basically only `semantics` would be needed fo assistance services. 
II also had a look at `Modifier.selectable` which is the seame  `clickable` + `semantics`

soulcramer (approved):

soulcramer (approved):

PR #417, opened 2023-05-30
chore: Update Icons and  Badges to support Color

MarinaRomanova:
## 📋 Changes description
- Update Icon APIs to accept `Color`so that it is possible to manage disabled state or animate color.
- Restore depricated `Badge`APIs for easier integration.
- Improve documentation.

## 📸 Screenshots
Available in this PR


spark-screenshot-testing/src/test/snapshots/images/com.adevinta.spark_PreviewScreenshotTests_preview_tests_appbar_bottomappbar_pro_dark.png
soulcramer, line 1: Isn't the default Icon size 16dp seing these changes?

soulcramer (commented):
The defautl size seems wrong as the icons gets too bing in the previews but otherwise LGTM

SimonMarquis:
> The defautl size seems wrong as the icons gets too bing in the previews but otherwise LGTM

If this is the Spark specs, so be it. We'll then check the diff on the apps and update appropriately.

SimonMarquis (approved):

PR #419, opened 2023-05-31
fix: Handle disabled colors for ButtonOutlined

### LINKED ISSUE #399: [Bug][Button] : Disabled button border

1.0.0

### 🐛 What happened?

When updating the "enabled" state of the button, the **border** of the button should be updated. 
That is currently not the case while using spark 0.1.0. 

For example : 

```
ButtonOutlined(
    onClick = {},
    enabled = uiModel.isEnabled,
    text =...,
    modifier = Modifier .fillMaxWidth()
)
```

And the result :

Screenshot 2023-05-24 at 16 59 35

### Relevant logcat output

_No response_

### Code of Conduct

MarinaRomanova:
## 📋 Changes description
- Update content color and border color of the `ButtonOutlined` when disabled. 

## 📸 Screenshots
Present in the current PR

## 🗒️ Other info
close #399

### DIFF
--- a/spark/src/main/kotlin/com/adevinta/spark/components/buttons/ButtonOutlined.kt
+++ b/spark/src/main/kotlin/com/adevinta/spark/components/buttons/ButtonOutlined.kt
@@ -30,6 +30,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.compositeOver
 import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.tooling.preview.Preview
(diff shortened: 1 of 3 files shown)


1ud0v1c (dismissed):

spark/src/main/kotlin/com/adevinta/spark/components/buttons/ButtonOutlined.kt
```diff
@@ -76,8 +77,12 @@ public fun ButtonOutlined(
         targetValue = intent.colors().color,
         label = "content color",
     )
+    val disabledContentColor = contentColor.copy(alpha = SparkTheme.colors.dim3)
+        .compositeOver(SparkTheme.colors.surface)
```
soulcramer, line 81: We don't want the composition on the `surface` here as the button may be displayed on other surfaces than `surface`

soulcramer (approved):

PR #437, opened 2023-06-08
feat: Add support for theming font family and setup Nunito as default

### LINKED ISSUE #415: Add support for downloadable fonts in the typography context
Currently we only use the default one which is Robotto but spark is using Nunito.
We need to either find a way to make spark work with downloadable font or set it up in LBC.

soulcramer:
## 📋 Changes description

Add an new param to `SparkTheme` to enable consumer to change the font family of their typography and allow the usage of downloadable fonts

## 🤔 Context

Currently we were using the default font which is Roboto. The issue is that the official font for spark is [Nunito](https://fonts.google.com/specimen/Nunito) and we needed to use 

### Roboto
image
### Nunito
image

## 🗒️ Other info

close #415


spark/src/main/res/values/font_certs.xml
```diff
@@ -0,0 +1,37 @@
...
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+  SOFTWARE.
+  -->
+<resources>
+    <string-array name="spark_com_google_android_gms_fonts_certs_dev">
+        <item>
```
soulcramer, line 26: @SimonMarquis I removed the cert string for now but is it okay to publish them in git?

gradle/libs.versions.toml
```diff
@@ -54,6 +54,8 @@ androidx-compose-material3-windowSizeClass = { module = "androidx.compose.materi
 androidx-compose-runtime = { module = "androidx.compose.runtime:runtime" }
 androidx-compose-runtime-livedata = { module = "androidx.compose.runtime:runtime-livedata" }
 androidx-compose-ui = { module = "androidx.compose.ui:ui" }
+# https://androidx.tech/artifacts/compose.ui/ui-util/
```
SimonMarquis, line 57: can we link to the officiel docs instead?
https://developer.android.com/reference/kotlin/androidx/compose/ui/text/googlefonts/package-summary

If they start to add written docs (like in [room](https://developer.android.com/reference/kotlin/androidx/room/package-summary) for instance) we would not be able to view them.

spark/src/main/kotlin/com/adevinta/spark/tokens/Font.kt
```diff
@@ -0,0 +1,83 @@
...
+import kotlinx.coroutines.CoroutineExceptionHandler
+
+
+public fun sparkFontFamily(
+    fontFamily: FontFamily = nunitoFontFamily,
+    useSparkTokensHighlighter: Boolean = false,
+    isLegacy: Boolean = false,
+    fontHandler: CoroutineExceptionHandler = defaultFontHandler,
+): SparkFontFamily {
+    return SparkFontFamily(
```
SimonMarquis, line 43: nit: return without method body

spark/src/main/kotlin/com/adevinta/spark/tokens/Font.kt
```diff
@@ -0,0 +1,83 @@
...
+import androidx.compose.ui.platform.LocalInspectionMode
+import androidx.compose.ui.text.font.FontFamily
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.googlefonts.Font
+import androidx.compose.ui.text.googlefonts.GoogleFont
+import com.adevinta.spark.R
+import kotlinx.coroutines.CoroutineExceptionHandler
+
+
+public fun sparkFontFamily(
```
SimonMarquis, line 37: If this must be part of the public API we should document it.

spark/src/main/kotlin/com/adevinta/spark/tokens/Font.kt
```diff
@@ -0,0 +1,83 @@
...
+internal val sparkFontProvider = GoogleFont.Provider(
+    providerAuthority = "com.google.android.gms.fonts",
+    providerPackage = "com.google.android.gms",
+    certificates = R.array.spark_com_google_android_gms_fonts_certs,
+)
+
+internal val nunitoFontFamily = FontFamily(
+    Font(googleFont = nunitoFont, fontProvider = sparkFontProvider, weight = FontWeight.Normal),
+    Font(googleFont = nunitoFont, fontProvider = sparkFontProvider, weight = FontWeight.SemiBold),
+    Font(googleFont = nunitoFont, fontProvider = sparkFontProvider, weight = FontWeight.Bold),
```
SimonMarquis, line 77: nit: can we add named parameters (or comments if this comes from java)

spark/src/main/kotlin/com/adevinta/spark/tokens/Font.kt
```diff
@@ -0,0 +1,83 @@
...
+)
+
+internal val defaultFontHandler = CoroutineExceptionHandler { _, throwable ->
+    // process the Throwable
+    Log.e("Font Resolver", "There has been an issue: ", throwable)
```
SimonMarquis, line 82: nit: I would remove the message `"There has been an issue: "` as it brings no value to the error message. Is there a `Log.e(TAG, throwable)`?

spark/src/main/kotlin/com/adevinta/spark/tokens/Type.kt
```diff
@@ -141,6 +141,26 @@ public fun sparkTypography(
     callout = callout,
 )
 
+/**
+ * Copy
+ */
```
SimonMarquis, line 146: Copy? 😕

spark/src/main/kotlin/com/adevinta/spark/tokens/Type.kt
```diff
@@ -141,6 +141,26 @@ public fun sparkTypography(
...
+@Composable
+internal fun SparkTypography.updateFontFamily(fontFamily: SparkFontFamily): SparkTypography {
+    return copy(
```
SimonMarquis, line 149: nit: Return without method body

spark/src/main/res/values/font_certs.xml
SimonMarquis: We should keep the hashes in this file. They are public values.

SimonMarquis (changes requested):

spark/src/main/kotlin/com/adevinta/spark/tokens/Font.kt
soulcramer: We cant add it here it seems since it's an vararg so I changed it to the list overload

spark/src/main/kotlin/com/adevinta/spark/tokens/Font.kt
```diff
@@ -33,27 +33,38 @@ import androidx.compose.ui.text.googlefonts.GoogleFont
 import com.adevinta.spark.R
 import kotlinx.coroutines.CoroutineExceptionHandler
 
-
+/**
+ * Creates a [SparkFontFamily] with the given [fontFamily] and [useSparkTokensHighlighter].
+ * @param fontFamily The [FontFamily] you want to use globally inside your app.
+ * @param useSparkTokensHighlighter If true, the [FontFamily] will be replaced by [FontFamily.Cursive] when the
+ * token highlighting is enabled in the SparkTheme.
+ * @param isLegacy If true, the [FontFamily] will be replaced by [FontFamily.Default] as we still need to support the legacy UI in LBC.
```
SimonMarquis, line 41:
```suggestion
 * @param isLegacy If true, the [FontFamily] will be replaced by [FontFamily.Default] as we still need to support the legacy UI.
```

spark/src/main/res/values/font_certs.xml
SimonMarquis: I forgot to send another message:

---

I think we should not rename these string resources, and keep the orignial ones:

`com_google_android_gms_fonts_certs` etc.

spark/src/main/res/values/font_certs.xml
```diff
@@ -0,0 +1,39 @@
...
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+  SOFTWARE.
+  -->
+<resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="ResourceName">
```
SimonMarquis, line 24:
```suggestion
<resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="ResourceName,Typos">
```

spark/src/main/res/values/font_certs.xml
SimonMarquis: No, why do you think that?

SimonMarquis (approved):

PR #373, opened 2023-05-05
feat: Update the catalog app design and add examples

soulcramer:
## 📋 Changes description

This PR try to add the support of sample of code usable in **Dokka** and in the catalog app for demo purposes.
I've also renamed the sample app into catalog app not avoid confusion between the samples of code and the demo app.

## 🗒️ Other info

close #361

Still in draft since the dokka task can't find the sample while we can find it in the catalog and inside AS

PR #438, opened 2023-06-08
feat: Add backdrop component to be used in sample app

soulcramer:
## 📋 Changes description

This is a copy of the Backdrop component from M2 as it's not yet available in M3.

## 🤔 Context

We made changes for the sample app and this is a part of the change in #373


catalog/src/main/kotlin/com/adevinta/spark/catalog/backdrop/Backdrop.kt
```diff
@@ -0,0 +1,542 @@
...
+import com.adevinta.spark.components.snackbars.SnackbarHostState
+import com.adevinta.spark.components.surface.Surface
+import com.adevinta.spark.tokens.contentColorFor
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.launch
+import kotlin.math.max
+import kotlin.math.min
+import kotlin.math.roundToInt
+
+// TODO-scott.rayapoulle.ext (23-15-2023): Remove once M3 has this implementation available
```
SimonMarquis, line 82: `23-15-2023`? :P

sample/src/main/kotlin/com/adevinta/spark/sample/backdrop/Backdrop.kt
```diff
@@ -0,0 +1,542 @@
...
+import com.adevinta.spark.components.snackbars.SnackbarHostState
+import com.adevinta.spark.components.surface.Surface
+import com.adevinta.spark.tokens.contentColorFor
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.launch
+import kotlin.math.max
+import kotlin.math.min
+import kotlin.math.roundToInt
+
+// TODO-scott.rayapoulle.ext (23-15-2023): Remove once M3 has this implementation available
```
SimonMarquis, line 82: nit: It could be helpful to keep track of the origin of this file. Do we have a link to the source?

SimonMarquis (changes requested):
Keep only one of the source files, either catalog/ or /sample

SimonMarquis (approved):

SimonMarquis (approved):

PR #439, opened 2023-06-08
chore: move / rename app from sample to catalog app

soulcramer:
## 📋 Changes description

Rename the sample app package from `sample` to `catalog` as it now contains more than just samples of components

## 🤔 Context

We made changes for the sample app and this is a part of the change in #373


SimonMarquis (dismissed):
LGTM 👍

SimonMarquis (dismissed):

SimonMarquis (approved):

PR #443, opened 2023-06-09
chore: Kotlin 1.8.22

SimonMarquis:
- Closes #436
- Closes #442


build-logic/src/main/kotlin/com/adevinta/spark/SparkAndroidComposePlugin.kt
```diff
@@ -39,6 +41,14 @@ internal class SparkAndroidComposePlugin : Plugin<Project> {
                 }
             }
 
+            @Suppress("ktlint:max-line-length")
```
soulcramer, line 44: Didn't we had this already 🤔 ?

soulcramer (dismissed):

build-logic/src/main/kotlin/com/adevinta/spark/SparkAndroidComposePlugin.kt
SimonMarquis: The `ktlint` suppression or the `suppressKotlinVersionCompatibilityCheck`?

SimonMarquis: this is not needed when the kotlin version matches the expected version requested by the compose compiler. And this is not the case here.
Compose compiler version 1.4.7 expects Kotlin 1.8.21.
This is not very important for such patch release, but we can wait until compose releases (if it ever does) a new 1.4.8 version.

soulcramer: Okay good to know thanks

SimonMarquis:
@soulcramer PTAL

soulcramer (approved):

PR #510, opened 2023-07-06
feat: Add extensions for applying dim to colors and weight to typos

### LINKED ISSUE #493: Add extension function to make usage of dims and highlight easier
[Context](https://adevinta.slack.com/archives/C04QQEU3Y2H/p1687164621601089)

 Having to right 
```kotlin
SparkTheme.typography.body2.copy(fontWeight = FontWeight.Bold)
``` 
or
```kotlin
SparkTheme.colors.primary.copy(alpha = SparkTheme.colors.dim3)
``` 
is becoming tedious even with the wrap Composable for emphasis.

One way we could solve this would be to add extension functions like :
```kotlin
@Composable
val Color.dime3 = copy(alpha = SparkTheme.colors.dim3)

Icon(
   color = SparkTheme.colors.primary.dim3,
)
```

MarinaRomanova:
close #493

### DIFF
--- a/spark/src/main/kotlin/com/adevinta/spark/tokens/Type.kt
+++ b/spark/src/main/kotlin/com/adevinta/spark/tokens/Type.kt
@@ -237,12 +237,12 @@ public data class SparkTypography(
     @Deprecated(
         "This property will be removed as it is not part of Spark Token",
         replaceWith = ReplaceWith(
-            expression = "body1.copy(fontWeight = FontWeight.Bold)",
+            expression = "body1.highlight",
             "androidx.compose.ui.text.font.FontWeight",
         ),
         level = DeprecationLevel.WARNING,
     )
-    val largeImportant: TextStyle = body1.copy(fontWeight = FontWeight.Bold)
+    val largeImportant: TextStyle = body1.highlight
 
     @Deprecated(
         "This property will be removed as it is not part of Spark Token",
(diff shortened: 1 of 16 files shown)


spark/src/main/kotlin/com/adevinta/spark/tokens/Type.kt
```diff
@@ -410,3 +410,6 @@ private fun TextPreview(
         )
     }
 }
+
+public val TextStyle.Highlight: TextStyle
```
SimonMarquis, line 414: Just wondering, is this `Highlight` term something defined in MaterialDesign/Spark? Because it has a totally different meaning to me.

SimonMarquis (approved):

spark/src/main/kotlin/com/adevinta/spark/tokens/Type.kt
MarinaRomanova: Yes, here is the link https://zeroheight.com/1186e1705/p/831208-typography/b/45b90c#:~:text=the%20jovial%20queen-,Body%201%20Highlight,-Family

MarinaRomanova: however, i think I will go for a lower case to have more consistency with other typo's namings

SimonMarquis (approved):

PR #517, opened 2023-07-10
fix: Fix MultilineTextField Alignment & TextField height

soulcramer:
## 📋 Changes description

Changed the TextField height to remove the different height when a supporting text was added and fixed the misalignment of addons when multiple lines were present in a TextField

## 🤔 Context

these bugs were found while testing the integration of the V0.3.0 in Polaris

## 📸 Screenshots

image

### DIFF
--- a/spark/src/main/kotlin/com/adevinta/spark/components/snackbars/Snackbar.kt
+++ b/spark/src/main/kotlin/com/adevinta/spark/components/snackbars/Snackbar.kt
@@ -94,7 +94,7 @@ public fun SparkSnackbar(
         modifier = modifier
             .padding(16.dp)
             .sparkUsageOverlay(),
-        shape = if(LocalLegacyStyle.current) SparkTheme.shapes.extraSmall else SparkTheme.shapes.small,
+        shape = if (LocalLegacyStyle.current) SparkTheme.shapes.extraSmall else SparkTheme.shapes.small,
         actionOnNewLine = actionOnNewLine,
         containerColor = colors.baseColor,
         contentColor = contentColor,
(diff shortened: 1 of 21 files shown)


spark/src/main/kotlin/com/adevinta/spark/components/snackbars/Snackbar.kt
```diff
@@ -94,7 +94,7 @@ public fun SparkSnackbar(
         modifier = modifier
             .padding(16.dp)
             .sparkUsageOverlay(),
-        shape = if (LocalLegacyStyle.current) SparkTheme.shapes.extraSmall else SparkTheme.shapes.small,
+        shape = if(LocalLegacyStyle.current) SparkTheme.shapes.extraSmall else SparkTheme.shapes.small,
```
soulcramer, line 97:
```suggestion
        shape = if (LocalLegacyStyle.current) SparkTheme.shapes.extraSmall else SparkTheme.shapes.small,
```

spark/src/main/kotlin/com/adevinta/spark/components/textfields/SparkTextField.kt
```diff
@@ -406,6 +403,9 @@ need to add additional padding themselves
 */
 internal val OutlinedTextFieldTopPadding = 8.dp
 
+// The dMaterial height is 56.dp with 16.dp vertical padding, since we want 12.dp for ours we subtract 8.dp
```
SimonMarquis, line 406:
```suggestion
// The Material height is 56.dp with 16.dp vertical padding, since we want 12.dp for ours we subtract 8.dp
```

PR #524, opened 2023-07-11
fix: Remove Small size for the button

### LINKED ISSUE #453: Remove Small size for button
We currently support the Small size for button but it require a custom size handling which force the component height.
This makes all buttons incompatible with font scaling so we need to remove it and go back to a simpler way to handle this (only apply padding based on the size by adding it to the font size?)

MarinaRomanova:
## 📋 Changes description

Remove Small size for the buttons as it does not respect accessibility guidelines and its usage was highly discouraged on mobile.

## 📸 Screenshots

Present in the current PR

## 🗒️ Other info

close #453


soulcramer:

soulcramer (approved):

PR #540, opened 2023-07-20
chore: Refactor checkbox group example

MarinaRomanova:
## 📋 Changes description
Refactor an example of checkbox group in CatalogApp
- Use loop to manage parent and children states.
In the case of a large group of checkboxes, it would be inefficient to create a state for each child checkbox.


mazzouzi (commented):
OK for me

PR #587, opened 2023-08-02
feat: Use dim3 for all components to handle disabled state

### LINKED ISSUE #575: Update disabled colors for all components
- make all components `enabled color`.dim 3 for disabled state instead of `container` color

MarinaRomanova:
## 📋 Changes description

Update
- Buttons
- Chips
- Tabs
- Toggles
to use `disabled`extension property that applies `dim3`composite overs `SparkTheme.colors.surface` to avoid unwanted transparency 

## 🤔 Context
Some of the components used
- dim3
- `container` colors
for disabled states. 
This refactoring allows to ensure the coherent styling

## 📸 Screenshots

Present in the current PR

## 🗒️ Other info

close #575


soulcramer (commented):
Are your sure about this change?
From what I understood from our meeting is that we would apply a disabled alpha to the whole component instead of composing it in some places 🤔

soulcramer:
Otherwise LGTM ping me again later

soulcramer (approved):

PR #588, opened 2023-08-02
fix: TextField padding between leadingContent and Label/BasicTextField

### LINKED ISSUE #582: [Textfield] The leading content has a end padding of `16.dp` while it should be `8.dp`

soulcramer:
## 📋 Changes description

We were using the default textfield horizontal spacing (16.dp) when even with leading/trailing content so we now use 8.dp only if there's one.

## 🤔 Context

As mentioned after a Design Review the spacing was bigger than wanted so we needed to reduce it to the correct value.

## 📸 Screenshots

image

Close #582


MarinaRomanova (approved):

soulcramer:

PR #589, opened 2023-08-03
fix: TextField background is modified generally and in disabled state

### LINKED ISSUE #583: [Textfield] Add background for disabled state

soulcramer:
## 📋 Changes description

The background is now Transparent to avoid overdraw in cases where the parent background is background or surface and in cases it isn't we won't have the square shown on the label caused by the clipping to remove the border.

Also we add a background color for disabled state as specified by the design.

## 📸 Screenshots

image

## 🗒️ Other info
Close #583


SimonMarquis (dismissed):
This looks very "dark" but LGTM otherwise 👍

MarinaRomanova (approved):

PR #594, opened 2023-08-04
feat: add the ability to change the launcher icon color depending on the theme mode

soulcramer:
## 📋 Changes description

Create color references and use them in the adaptive icon foreground & background drawables to change the icon color on the theme mode

## 🤔 Context

I was inspired by [this PR on NowInAndroid](https://github.com/android/nowinandroid/pull/809) as I didn't knew we could change the color of the icon depending on the theme 🤩


SimonMarquis (approved):
Nice ✨

PR #598, opened 2023-08-08
fix: remove trailing slash from `Url`s

SimonMarquis:
Most of them are used inside String templating, and would introduce unexpected double slashes that could break the url.

For example:
- ✅ https://adevinta.github.io/spark-android/spark/com.adevinta.spark.components.appbar/index.html
- ❌ https://adevinta.github.io/spark-android/spark//com.adevinta.spark.components.appbar/index.html

### DIFF
--- a/catalog/src/main/kotlin/com/adevinta/spark/catalog/util/Url.kt
+++ b/catalog/src/main/kotlin/com/adevinta/spark/catalog/util/Url.kt
@@ -31,17 +31,17 @@ public fun Context.openUrl(url: String) {
 }
 
 public const val GuidelinesUrl: String = "https://spark.adevinta.com"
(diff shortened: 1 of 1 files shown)


soulcramer (commented):
LGTM but I still need to check if this doesn't break the links for zeroheight

SimonMarquis:
- https://spark.adevinta.com/1186e1705 works fine
- and composed urls as well

https://github.com/adevinta/spark-android/blob/bb5c7bf72e660d36014ac7f7e94dd8bb1a2557d1/catalog/src/main/kotlin/com/adevinta/spark/catalog/model/Components.kt#L60

kazaky (approved):
LGTM

PR #604, opened 2023-08-09
fix: content was not adapting to button spaces when too big

### LINKED ISSUE #599: [FullScreenModal] Bug: Incorrect content scale of illustration
## Illustration 
🐛 The illustration is croped instead to fit the given width
💄The size of the illustration can be augmented.

## In Landscape orientation on mobile 
 - the image should be centered
-  the buttons stay visible

soulcramer:
## 📋 Changes description

- Make the content take a weight with fill false to reduce its size priority so that it doesn't push the buttons out of the screen
- Update close icon color to `onSurface.dim2` and add content description for a11y
- Allow passing `illustrationContentScale` to handle content scale of the illustration (the default is `ContentScale.Fit`)
- Add surface to the buttons and update padding so that it is spaced with the content
- Add padding for dialog window to avoid it touching the screen edges

close #599


MarinaRomanova (dismissed):

MarinaRomanova:
@spark-ui-bot spotlessApply

kazaky (approved):
LGTM

PR #630, opened 2023-08-17
feat: Update front layer content surface color

MarinaRomanova:
## 📋 Changes description
- Use onSurface.dim5 as a content colors for front layer
- update card used for Showkase part

## 🤔 Context
Before Surface color was used as a front layer content color, however such components as ButtonContrast cannot be properly seen as they also use Surface as the background color.

## 📸 Screenshots
Before: 
 

After:


MarinaRomanova:
SurfaceVariant color is depricated though

MarinaRomanova:
Screenshot_1692349084
I tried to play with backgroundVariant color and adding it solely to the sample component, but i find it less "readable" and also it should be added for each component every time they have Contrast style

PR #648, opened 2023-08-22
chore: Fetch icon resources in coroutines in CatalogApp

MarinaRomanova:
## 📋 Changes description
We have 400+ icons at the moment to avoid freezing UI get all icons in couroutines.


SimonMarquis (approved):
LGTM

PR #671, opened 2023-08-24
chore: prepare next development version

soulcramer:


SimonMarquis:
I dont think we need this commit. Simply merge the tag reference on main and keep the current version in `gradle.properties`.

PR #674, opened 2023-08-25
chore: Add IconButton screenshot test

MarinaRomanova:
## 📋 Changes description
- Add Screenshot test for IconButtons
- Remove previews of IconButtons for non-default sizes

## 🤔 Context

We have agreed to add screnshot test with Paparazzi rather than adding numerous Previews. 
- Configurator part in CatalogApp is meant to be used to see the component with all available properties. 
- Paparazzi should be used for regression tests 

## 📸 Screenshots
Present in the current PR


SimonMarquis:
Do we really need this many images? 🙈
Is there a way we can merge things?

MarinaRomanova:
I refactored the grouping, we have much less now 😁

PR #703, opened 2023-09-05
chore: Remove Marina as code owner

soulcramer:
## 📋 Changes description

Remove marina from the code owner file to avoid sending a ping on every PR

## 🤔 Context

Since Marina is no longer a member of the org she shouldn't be considered a code owner anymore


soulcramer:
🫡

turolopezsanabria:
😢

PR #740, opened 2023-09-25
fix: Fix outline & ghost style icon button bug

kazaky:
## 📋 Changes description

In catalog app, for ghost and outlined styles there's a white background behind the iconButton (for all intents, and loading & not loading)

## 🤔 Context
This PR solves this issue

## ✅ Checklist

- [x] Link to GitHub issues it solves. closes #726 
- [x] I have reviewed the submitted code.
- [x] I have tested on a phone device/emulator.

## 📸 Screenshots
Screenshot 2023-09-25 at 4 06 30 AM

## 🗒️ Other info

Early there was a PR https://github.com/adevinta/spark-android/pull/702
to introduce material 2 elevation instead of material 3 elevation

Using M2 Surface caused this bug as the surface is being drawn even if container color from color intent is transparent color,
**expected behavior:**
Background Surface should not be drawn if container color is transparent

[Contributing](https://github.com/adevinta/spark-android/blob/main/docs/contributing.md) has more information and tips for a great pull request.

PR #744, opened 2023-09-25
fix: Other APi of surface where using the wrong elevation

### LINKED ISSUE #726: [Bug] IconButton - fix ghost and outlined styles
In catalog app, for `ghost` and `outlined` styles there's a white background behind the iconButton (for all intents, and loading & not loading)

isolated

Video of the bug here ⬇ 
https://user-images.githubusercontent.com/142396647/268904873-a35b1906-755b-4543-ad5c-6629553bf2de.mov

soulcramer:
## 📋 Changes description

Use elevation instead of absolute elevation as shadow elevation in `Surface`.

## 🤔 Context

This changed was inspired from #740 
In #702 we updated the `Surface` to use the M2 elevation system instead of the tinted one from M3. During the dev I made the mistake of using the absolute elevation as shadow elevation.

This was fixed but I found out it was only in one of the 4-5 different api for `Surface`. As we use these api for icon buttons every buttons where using the absolute elevation and was more visible on buttons with no background like `GhostButton`

close #726

## ✅ Checklist

## 📸 Screenshots

image

### DIFF
--- a/spark/src/main/kotlin/com/adevinta/spark/components/surface/Surface.kt
+++ b/spark/src/main/kotlin/com/adevinta/spark/components/surface/Surface.kt
@@ -231,7 +231,7 @@ public fun Surface(
                         absoluteElevation = absoluteElevation,
                     ),
                     border = border,
-                    elevation = absoluteElevation,
+                    elevation = elevation,
                 )
                 .clickable(
                     interactionSource = interactionSource,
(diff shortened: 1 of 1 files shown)


SimonMarquis (approved):

kazaky (approved):
LGTM

PR #748, opened 2023-09-27
feat(ci): extract common setup code in composite actions

SimonMarquis:
And remove unused `gradle.yml` workflow.


soulcramer (approved):
LGTM though can you tell me why we used these versions of actions before? I can't remember it

SimonMarquis:
> why we used these versions of actions before

What do you mean? I don't understand 🤔

PR #749, opened 2023-09-27
feat(ci): restore support for merge queue behavior

SimonMarquis:
- Updating the concurrency group to include the merge ref.
- Removing the custom analysis path (to be checked later).


kazaky (approved):
LGTM

PR #767, opened 2023-10-06
fix: checkbox checkmark was using onPrimary instead of onColor from intent

soulcramer:
## 📋 Changes

Use intent on color for checkbox checkmark and use `Transparent` as surface color in configurators when the intent is not `surface`

## 🤔 Context

When testing the KA flavor of Polaris I noticed that the checkbox was using `onMain` for the checkmark which is the same color as basic in this theme and would make the checkbox fully basic.

## ✅ Checklist

- [x] I have tested on a phone in the catalog app.

## 📸 Screenshots

image

image


SimonMarquis (approved):
LGTM 👍

SimonMarquis:
@spark-ui-bot spotless

SimonMarquis:
🙈 spotless command will be fixed in https://github.com/adevinta/spark-android/pull/768

kazaky (approved):
LGTM

soulcramer:

PR #772, opened 2023-10-11
feat: rely on default ktlint version provided by spotless (1.0.0)

SimonMarquis:


SimonMarquis:
@soulcramer & @kazaky can you both confirm if you agree on this configuration?

soulcramer (approved):
LGTM, this is nicer than before thanks 👍

kazaky (approved):
LGTM

PR #782, opened 2023-10-16
chore(gradle): Simpler Kotlin DSL Property Assignment

SimonMarquis:
https://blog.gradle.org/simpler-kotlin-dsl-property-assignment


soulcramer (approved):
Nice :o

PR #798, opened 2023-11-07
feat: Add Subito theme to catalog app

soulcramer:
## 📋 Changes

Replace The legacy LBC theme with subito new theme and update the color visibility so that each theme use its own colors

## 🤔 Context
Subito has finished building their theme 

## 📸 Screenshots

image
image


soulcramer:
Yep we don't test it anymore in Spark, it's not yet removed in Polaris though 👍

kazaky (approved):
LGTM

PR #815, opened 2023-11-20
fix: Replace zeroheight links with spark.adevinta.com

SimonMarquis:
## 🤔 Context

Now that we have a proper domain name, we should use it 🤓


spark/src/main/kotlin/com/adevinta/spark/components/textfields/SelectTextFields.md
```diff
@@ -1,6 +1,6 @@
 # Package com.adevinta.spark.components.textfields
 
-[SelectTextFields](https://zeroheight.com/25c15666f/p/29d201-textfield-) allow users to select an
+[SelectTextFields](https://spark.adevinta.com/1186e1705/p/773c60-input--text-field/b/0658e2) allow users to select an
```
SimonMarquis, line 3: nit: there are no proper link for `SelectTextFields`, so I used the main `Input / Text-field` component.

soulcramer (approved):

PR #874, opened 2024-01-03
fix: Add Info as Icon buttons intent color

### LINKED ISSUE #706: Buttons are missings the info intent color
The buttons were made based on the Specs & Handhoff file but it's not updated since the color refactor.

On this file we can see that this intent is not present but it's available to designers in the Components file so we need to add it.
cc @elisa-hery

soulcramer:
## 📋 Changes

IconButton and IconToggleButton were missing the info intent color

## 🤔 Context

I think this was because this intent was added later to the button intents

Close #706 

## ✅ Checklist

## 📸 Screenshots

image


SimonMarquis (approved):

SimonMarquis:
@soulcramer btw, I feel like the `Style` segmented buttons should be split into something else, as we now have 5 different styles, and it makes them ellipsised.

soulcramer:
5 is the max number of options available for a Segmented Button here the labels are too big (ideally it would be just numbers or icons for the number of options) AND since I'm using a bigger font scaling than the default one it doesn't help haha 😂

PR #880, opened 2024-01-10
fix: Attempt to reduce letter spacing on callout typo

soulcramer:
## 📋 Changes

Enforce the usages of the typo callout in button even if it's was already the right one used to avoid regressions on refactor and specify explicitly the `0.sp` letter spacing on typo to ensure that it's the value used everywhere.

## 🤔 Context

It seems like the typo in buttons fro android are "bigger" so after looking into it I didn't found any clue if this was really the case or not so I'm explicitly configuring the typo in case the defaults are not what we want.

## ✅ Checklist

## 📸 Screenshots

Can be found in changed files


kazaky (approved):
LGTM

PR #946, opened 2024-02-07
fix: remove unexpected Compose tooling dependency on runtime classpath

SimonMarquis:


soulcramer (approved):
Thanks for the change, I saw last year that they splited the modifiers from the tooling but forgot about it 😬

PR #947, opened 2024-02-07
chore: Update modifiers impl to use Node api and suppress Compose lint warnings on forked components

soulcramer:
## 📋 Changes

Make the following `sparkUsageOverlay` modifier to use the Modifier.Node api, the modifier `minimumTouchTargetSize` to use the material one as it has been made public and suppress the coming warnings from the next version of compose lint  for components that we forked

## 🤔 Context

As said in https://github.com/adevinta/spark-android/pull/923#issuecomment-1924447475 this PR aims to surface the changes required to upgrade to the next compose lint version

## ✅ Checklist


spark/src/main/res/values-fr/strings.xml
```diff
@@ -33,12 +33,12 @@
     <plurals name="spark_rating_with_comments_a11y" tools:ignore="UnusedQuantity">
         <item quantity="one">Note de %1$.1f pour %2$d avis</item>
         <item quantity="other">Note de %1$.1f pour %2$d avis</item>
-        <item quantity="many">Note de %1$.1f pour %2$d avis</item>
+        <item quantity="many">Note de %1$.1f pour %2$d d’avis</item>
```
SimonMarquis, line 36: Are you sure about these additions of `d'` and `de`?

spark/src/main/kotlin/com/adevinta/spark/tools/modifiers/TouchTarget.kt
```diff
@@ -34,40 +34,57 @@ import androidx.compose.ui.unit.dp
...
 
 /**
- * Ensure that the composable has a minimum touch target size
+ * Reserves at least 44.dp in size to disambiguate touch interactions if the element would measure
+ * smaller.
  *
- * Duplicate of Material minimumTouchTargetSize() since it's internal on their side
+ * https://m3.material.io/foundations/accessible-design/accessibility-basics#28032e45-c598-450c-b355-f9fe737b1cd8
+ *
+ * This uses the Spark recommended minimum size of 44.dp x 44.dp, which may not the same as the
```
SimonMarquis, line 42:
```suggestion
 * This uses the Spark recommended minimum size of 44.dp x 44.dp, which may not be the same as the
```

spark/src/main/kotlin/com/adevinta/spark/tools/modifiers/TouchTarget.kt
```diff
@@ -34,40 +34,57 @@ import androidx.compose.ui.unit.dp
...
+ * system enforced minimum size. The minimum clickable / touch target size (44.dp by default) is
+ * controlled by the system via ViewConfiguration` and automatically expanded at the touch input layer.
```
SimonMarquis, line 44:
```suggestion
 * controlled by the system via ViewConfiguration and automatically expanded at the touch input layer.
```

spark/src/main/res/values-fr/strings.xml
soulcramer: Yeah I [checked french plural rules](https://www.unicode.org/cldr/charts/44/supplemental/language_plural_rules.html#fr) and for the many category it should/must be present

PR #1123, opened 2024-05-22
chore!: Remove the legacy API

### LINKED ISSUE #971: Remove legacy themes mentions from the tokens and components
During the Compose Migration of Polaris I've added a CompositionLocal that will apply different styles depending on it to respect the previous UI we had in Polaris and avoid 2 different style from being visible on a screen.

The migration being close to the end we can now safely remove this flags from our codebase

```[tasklist]
### Tasks
- [x] Remove `LocalLegacyStyle` from the code base and only use the current style where it was used
- [x] Remove the legacy theme from the catalog app
- [x] Remove the deprecated `TextButton`
- [x] Remove the deprecated `AssistChip`, `FilterChip`, `InputChip`, `SuggestionChip`
- [x] Remove the deprecated `TagPromote`, `TagUrgent`, `TagPro`
```

soulcramer:
## 📋 Changes

We remove all apis and mentions of the legacy Brikke theme / DS

## 🤔 Context

Closes #971 

## ✅ Checklist


catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Adevinta
+ * Copyright (c) 2023-2024 Adevinta
```
SimonMarquis, line 2: @soulcramer can you check your IDE if you don't have any custom copyright code, because this is not the expected format 🙈

soulcramer: Damn why is this coming back 💀
I'll check

kazaky (approved):

kazaky (dismissed):

PR #1181, opened 2024-06-20
feat(dropdown): Add multi selection example

### LINKED ISSUE #1162: [Dropdown] Showcase Multiselection
There may be places where the multi selection will be asked.
For this use case like for the Chips we will provide examples on how users can achieve this with this component instead of forcing a single implementation.

soulcramer:
## 📋 Changes

Add multi selection example
Also add the missing padding in the dropdown menu

## 🤔 Context

Closes #1162 

## ✅ Checklist

### DIFF
--- a/catalog/src/main/kotlin/com/adevinta/spark/catalog/examples/samples/text/DropdownExamples.kt
+++ b/catalog/src/main/kotlin/com/adevinta/spark/catalog/examples/samples/text/DropdownExamples.kt
@@ -22,17 +22,20 @@
 package com.adevinta.spark.catalog.examples.samples.text
 
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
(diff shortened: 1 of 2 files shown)


catalog/src/main/kotlin/com/adevinta/spark/catalog/examples/samples/text/DropdownExamples.kt
```diff
@@ -74,15 +77,91 @@ public val DropdownsExamples: List<Example> = listOf(
...
+                                onClick = {
+                                    selectedItems = if (book in selectedItems) {
+                                        selectedItems - book
+                                    } else {
+                                        selectedItems + book
+                                    }
+                                    // Here we want to have the dropdown to stay expanded when we multiSelect but
+                                    // you could use the line below if in the case you want to have the same behaviour
+                                    // as the single selection.
+                                    // expanded = false
```
kazaky, line 163: I would suggest to extract this part, to make the example more concise, easy to go through the code

kazaky (changes requested):
kindly check the comment

kazaky (approved):

soulcramer:
@kazaky Please don't merge PR's if they're not based on `main` otherwise it'll break the branch chain

PR #1273, opened 2024-09-05

SimonMarquis:


soulcramer (approved):
Icon color is changed since we now use spark color tokens

PR #1337, opened 2024-10-08
chore: upgrade slack lint and agp version

soulcramer:
## 🤔 Context

We need the version upgrade from agp and lint to be able to use the new version of slack lints and we need to upgrade it because the new versions of agp & lint removed an api used by the previous version of slack lint so we need to upgrade all of the mat the same time 😵‍💫


spark/lint.xml
```diff
@@ -22,11 +22,10 @@
   SOFTWARE.
   -->
 <lint>
-    <issue id="ComposeCompositionLocalGetter" severity="warning">
+    <issue id="ComposeCompositionLocalUsage" severity="warning">
         <option name="allowed-composition-locals"
             value="LocalSparkColors,LocalSparkShapes,LocalSparkTypography,LocalSparkFeatureFlag,LocalElevationOverlay,LocalWindowSizeClass" />
     </issue>
-    <issue id="AutoboxingStateCreation" severity="warning"/>
```
SimonMarquis, line 29: What is the reason behind this removal?

soulcramer: Mistakenly removed 😬

SimonMarquis (dismissed):

PR #1355, opened 2024-10-16
feat(Image): Add configurator

soulcramer:
## 📋 Changes

Add configurator with the shape graphic lib to demonstrate how to clip to custom shapes

## 🤔 Context

Close [SPA-65](https://jira.ets.mpi-internal.com/browse/SPA-65)

## ✅ Checklist

## 📸 Screenshots

image


catalog/src/main/kotlin/com/adevinta/spark/catalog/configurator/samples/image/ImageConfigurator.kt
```diff
@@ -0,0 +1,332 @@
...
+@Composable
+private fun ColumnScope.ImageSample() {
+    var state by remember { mutableStateOf(ImageState.Loading) }
+    var width by remember { mutableStateOf<Int?>(1) }
+    var height by remember { mutableStateOf<Int?>(1) }
+    var aspectRatio by remember { mutableStateOf(ImageAspectRatio.Custom) }
+    var imageShape by remember { mutableStateOf(ImageShape.Wavy) }
+    val painter =
+        rememberDrawablePainter(getDrawable(LocalContext.current, R.drawable.notifications))
+    val drawable = getDrawable(LocalContext.current, R.drawable.notifications)!!
```
SimonMarquis, line 93: Do we need to query getDrawable() twice?

spark/src/main/kotlin/com/adevinta/spark/components/textfields/SparkTextField.kt
```diff
@@ -230,7 +230,6 @@ internal fun SparkTextField(
                 modifier
             }
                 .defaultMinSize(
-                    minWidth = TextFieldDefaults.MinWidth,
```
SimonMarquis, line 233: Unrelated change to be removed

spark/src/main/kotlin/com/adevinta/spark/components/image/Image.kt
```diff
@@ -66,7 +66,7 @@ import com.adevinta.spark.tools.modifiers.sparkUsageOverlay
 
 @InternalSparkApi
 @Composable
-internal fun SparkImage(
+public fun SparkImage(
```
SimonMarquis, line 69: Can we keep it internal and add a flag to use it in the configurator?

soulcramer: Oh I was wandering how we can do it without triggering a compiler warning 🤔 
Do you have another solution than https://publicobject.com/2024/01/30/internal-visibility/ ?

I tried to use it but unfortunatly the compiler give me this warning

> This code uses error suppression for 'INVISIBLE_REFERENCE'. While it might compile and work, the compiler behavior is UNSPECIFIED and WON'T BE PRESERVED. Please report your use case to the Kotlin issue tracker instead: https://kotl.in/issue

SimonMarquis: This was the solution I was going to give you. There might be additional compiler options to turn this warning off.
Otherwise, you can try the Annotation that restrict usage to a specific library (I think AndroidX uses this a lot, but I don't remember the name of this annotation).

soulcramer: Should be resolved in e46e6116208ab87c0a0372dab5223b1689751000

PR #1377, opened 2024-11-11
feat(API): bump compileSdk & targetSdk to API level 35 (Android 15)

### LINKED ISSUE #1368: [FR]: Bump targetSdk and compileSdk to 35 (Android 15)

### Describe the problem

Spark is currently using API level 34

https://github.com/adevinta/spark-android/blob/42d16dfe8c9fea36cd368b50b631d881f7ff98e1/gradle/libs.versions.toml#L13

https://github.com/adevinta/spark-android/blob/42d16dfe8c9fea36cd368b50b631d881f7ff98e1/gradle/libs.versions.toml#L28

### Describe the solution

Update to API level 35, and check edge-to-edge system enforcement potential side-effects.

### Additional context

_No response_

### Code of Conduct

SimonMarquis:
## 📋 Changes

Bump `compileSdk` & `targetSdk` to API level 35 (Android 15).

## 🤔 Context

⚠️ Deprecated APIs: https://developer.android.com/about/versions/15/behavior-changes-15#deprecated-apis

Closes #1368


soulcramer (approved):

spark/src/main/kotlin/com/adevinta/spark/components/dialog/ModalScaffold.kt
```diff
@@ -137,6 +137,7 @@ public fun ModalScaffold(
                 )
     val activityWindow = getActivityWindow()
 
+    @Suppress("DEPRECATION")
```
SimonMarquis, line 140: @soulcramer do we need to do something about these APIs being used while deprecated? Are they still working as expected?

soulcramer: I checked and without them it works as expected with edge-to-edge and without 🤷

SimonMarquis: On all API levels you mean? (pre&post forced edge-to-edge)

soulcramer: yes I tested it on my emulator on 31 but I did not tested it without gestures, I'll try with that

soulcramer: I changed the Modal Scaffold API here https://github.com/adevinta/spark-android/pull/1393 to optin the edge to edge support, on early version it didn't work as expected with the navigation buttons where the bottom bar would draw behind it instead of on top of it

soulcramer (approved):

PR #1401, opened 2024-12-12
refactor: Use the new Typesafe routes for navigation

soulcramer:
## 📋 Changes

Replace all the string routes by typed ones.
Encapsulate the navigation to be only accesible from specific destinations/graphs

## ✅ Checklist

### DIFF
--- a/catalog/src/main/AndroidManifest.xml
+++ b/catalog/src/main/AndroidManifest.xml
@@ -43,6 +43,15 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data android:scheme="https" />
+                <data android:host="android.spark.adevinta.com" />
+            </intent-filter>
         </activity>
 
     </application>
--- a/catalog/src/main/AndroidManifest.xml
+++ b/catalog/src/main/AndroidManifest.xml
@@ -43,15 +43,6 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
-
-                <category android:name="android.intent.category.DEFAULT" />
-                <category android:name="android.intent.category.BROWSABLE" />
-
-                <data android:scheme="https" />
-                <data android:host="android.spark.adevinta.com" />
-            </intent-filter>
         </activity>
 
     </application>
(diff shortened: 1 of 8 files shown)


catalog/src/main/AndroidManifest.xml
```diff
@@ -43,6 +43,15 @@
...
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data android:scheme="https" />
+                <data android:host="android.spark.adevinta.com" />
+            </intent-filter>
```
SimonMarquis, line 54: I don't think this was added on purpose 🤔

catalog/src/main/kotlin/com/adevinta/spark/catalog/configurator/Navigation.kt
```diff
@@ -60,5 +58,6 @@ internal fun NavGraphBuilder.navGraph(
     }
 }
 
-internal const val ConfiguratorRoute = "configurator"
-internal const val ComponentIdArgName = "componentId"
+private fun NavController.navigateToConfiguratorShowcase(id: Int) {
+    navigate(route = ConfiguratorShowcase(componentId = id))
+}
```
SimonMarquis, line 63: Do we need to extract this single usage into a function?
In other parts of the code, you don't extract them.
We should choose one consistent way of doing it. Personnaly, I would inline these at the call site directly.

catalog/src/main/kotlin/com/adevinta/spark/catalog/examples/Navigation.kt
```diff
@@ -22,69 +22,67 @@
...
-internal const val ExampleRoute = "example"
-internal const val ComponentIdArgName = "componentId"
-internal const val ExampleIndexArgName = "exampleIndex"
+private fun NavController.navigateToExampleSection(id: Int) {
+    navigate(route = ExampleSection(componentId = id))
+}
+
+private fun NavController.navigateToExampleShowcase(id: Int, index: Int) {
+    navigate(route = ExampleShowcase(componentId = id, exampleIndex = index))
+}
```
SimonMarquis, line 88: Same here

catalog/src/main/AndroidManifest.xml
soulcramer: Ah actually it was but not directly but for a follow-up PR to add deep link support.
I'll readd it later

catalog/src/main/kotlin/com/adevinta/spark/catalog/examples/Navigation.kt
soulcramer: Fixed in https://github.com/adevinta/spark-android/pull/1401/commits/b231b1cc07ea384f7c81c9a628d7d93e900eb9a1

catalog/src/main/kotlin/com/adevinta/spark/catalog/configurator/Navigation.kt
soulcramer: Fixed in https://github.com/adevinta/spark-android/pull/1401/commits/b231b1cc07ea384f7c81c9a628d7d93e900eb9a1

PR #1402, opened 2024-12-12
feat: Add custom accessibility annotations to AnnotatedStringResource

francoisadam:
## 📋 Changes

- Added string annotation handling for all TtsSpan types inside the AnnotatedStringResource
- Convert strings with these annotations to more human readable equivalent
- Added a new tab in the catalog app to preview and check new accessibility improvements
- Added coreLibDesugaring to Spark to handle date parsing (will have to check for size impacts)

## 🤔 Context

For most content descriptions, TalkBack won't read properly the words unless we do some formatting.
Before Compose, there was the option to use TtsSpan to add metadate to text and help TalkBack properly read the words.
But there is only one annotation available for Compose as of right now (Verbatim), so we need to do it manually for now.

## ✅ Checklist

- [X] Add custom Tts annotations
- [X] Add a new tab in the catalog app to display and test these annotations
- [ ] Add unit tests once these new annotations are approved
- [ ] Add other annotations for specific strings like zipcode for example
- [ ] Put these accessibility annotation handling in a different extension to make sure this is only done for contentDescriptions and not for user-visible texts
- [ ] Make it i18n compatible ?


spark/src/main/kotlin/com/adevinta/spark/res/AnnotatedStringResource.kt
```diff
@@ -291,16 +297,25 @@ private fun CharSequence.asAnnotatedString(
...
             val start = getSpanStart(it)
             val end = getSpanEnd(it)
-            buildWithSpan(it, start, end, density, colors, typography)
+            annotatedString = annotatedString.buildWithAccessibilitySpan(it, start, end)
+        }
+        append(annotatedString)
+        getSpans(0, length, Any::class.java).forEach {
+            val start = getSpanStart(it)
+            val end = getSpanEnd(it)
+            buildWithSpanStyle(it, start, end, density, colors, typography)
```
SimonMarquis, line 310: I'm not sure to fully understand why we duplicated this part of the code. Can't we merge the tts handling with the already existing `buildWithSpan` which is already parsing the entire string/annotations once?

francoisadam: That's the frustrating part, TtsSpan in spannable uses metadata to add info for Tts to better read the words but this is not possible yet in compose and AnnotatedStrings (well only Verbatim is possible for now)

So the only way I found was to actually reformat the string itself to make it more readable by Tts.

The first part with `buildWithAccessibilitySpan` updates the strings before using `append`, and the second part with `buildWithSpanStyle` only adds styling to the annotated string (and so needs to be done after the `append`)

But if you know how to extend the TtsAnnotation from Compose, or a better way to do this I'm all ears 🙏

PR #1462, opened 2025-01-31
feat: add support for language selection

soulcramer:
## 📋 Changes

Add support for app language for french, english.

## 🤔 Context

Apply the documentation guide from 
https://developer.android.com/guide/topics/resources/app-languages#auto-localeconfig

This should make it easier to test with different locale without having to change the phone one every time.

## ✅ Checklist

## 📸 Screenshots

image
image
image


catalog/src/main/res/xml/locales_config.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+  Copyright (c) 2024 Adevinta
```
SimonMarquis, line 4: nit: 2025 😛

catalog/build.gradle.kts
```diff
@@ -31,6 +31,9 @@ plugins {
 android {
     namespace = "com.adevinta.spark.catalog"
     defaultConfig.applicationId = "com.adevinta.spark.catalog"
+    defaultConfig.resourceConfigurations.addAll(
+        setOf("en-rGB", "fr", "de"),
```
SimonMarquis, line 35: I'm not sure how en-us is visible on the list while not being present in this list 🤔

SimonMarquis: Also, I don't see any en-gb string resources in the codebase.

soulcramer: After searching around the "English (United States)" is suggested because it's "popular" 🤷
If a user select it, he will still use the en-GB since it's the only english local available (in this context)

catalog/src/main/res/values-en-rGB/strings.xml
```diff
@@ -0,0 +1,147 @@
...
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+  SOFTWARE.
+  -->
+<resources xmlns:tools="http://schemas.android.com/tools" tools:locale="en-GB">
```
SimonMarquis, line 24: Are these different from the base en locale? I feel like we should only provide overrides where it makes sense instead of duplicating everything. It will be easier to maintain.

catalog/src/main/res/values/strings.xml
```diff
@@ -21,7 +21,7 @@
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
   SOFTWARE.
   -->
-<resources xmlns:tools="http://schemas.android.com/tools" tools:locale="en">
+<resources xmlns:tools="http://schemas.android.com/tools" tools:locale="en-GB">
```
SimonMarquis, line 24: I don't get it. We are adding a new file for British English specifically, but we are still changing the default to the same locale.

soulcramer: I've removed it, I thought this impacted the suggested language but no.

SimonMarquis: Ok, but I don't think this is how it works. This is only a tools attribute, meaning it probably has no impact on the app itself.
When trying to load the specific en-gb locale, it will not find any, and will fallback to the default one.
That means the en-gb locale does not add anything but an extra entry in the locale list. Or am I missing something?

soulcramer: I didn't remove the tools label but the new (extra) values-en-rGB, I wasn't explicit sorry.
You're correct this only help for lint in this file and doesn't help at all for resolving which locale is available

PR #1473, opened 2025-02-14
feat(catalog): Add support for app deeplink

soulcramer:
## 📋 Changes

Add for custom scheme & host deeplink for all screens in the catalog app

## 🤔 Context

First step before adding applinks to allow deeplink from a url scheme

## ✅ Checklist

### DIFF
--- a/catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
+++ b/catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
@@ -332,4 +332,4 @@ private val lightScrim = android.graphics.Color.argb(0xe6, 0xFF, 0xFF, 0xFF)
  */
 private val darkScrim = android.graphics.Color.argb(0x80, 0x1b, 0x1b, 0x1b)
 
-internal val AppBasePath = "spark://catalog"
+internal const val AppBasePath = "spark://"
--- a/catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
+++ b/catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
@@ -179,10 +179,10 @@ internal fun ComponentActivity.CatalogApp(
 
         DeviceConfigurationOverride(
             override = DeviceConfigurationOverride.DarkMode(useDark)
-                then DeviceConfigurationOverride.LayoutDirection(layoutDirection)
-                then DeviceConfigurationOverride.FontScale(
-                    theme.takeUnless { it.fontScaleMode == System }?.fontScale ?: LocalDensity.current.fontScale,
-                ),
+                    then DeviceConfigurationOverride.LayoutDirection(layoutDirection)
+                    then DeviceConfigurationOverride.FontScale(
+                theme.takeUnless { it.fontScaleMode == System }?.fontScale ?: LocalDensity.current.fontScale,
+            ),
         ) {
             Box(
                 modifier = Modifier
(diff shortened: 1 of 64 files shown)


catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
```diff
@@ -207,8 +208,17 @@ internal fun ComponentActivity.CatalogApp(
             ) {
                 val coroutineScope = rememberCoroutineScope()
                 val homeScreenValues = CatalogHomeScreen.entries
+                val initialPage = LocalActivity.current?.intent?.data?.let {
+                    when {
+                        it.pathSegments.contains("examples") -> CatalogHomeScreen.Examples.ordinal
+                        it.pathSegments.contains("configurator") -> CatalogHomeScreen.Configurator.ordinal
+                        it.pathSegments.contains("icons") -> CatalogHomeScreen.Icons.ordinal
```
SimonMarquis, line 215: nit:
```suggestion
                        "examples" in it.pathSegments -> CatalogHomeScreen.Examples.ordinal
                        "configurator" in it.pathSegments -> CatalogHomeScreen.Configurator.ordinal
                        "icons" in it.pathSegments -> CatalogHomeScreen.Icons.ordinal
```

catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
```diff
@@ -207,8 +208,17 @@ internal fun ComponentActivity.CatalogApp(
             ) {
                 val coroutineScope = rememberCoroutineScope()
                 val homeScreenValues = CatalogHomeScreen.entries
+                val initialPage = LocalActivity.current?.intent?.data?.let {
```
SimonMarquis, line 211: IMO `initialPage` should be `CatalogHomeScreen`, and the call to `.ordinal` should be in the `rememberPagerState` call only.

catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
```diff
@@ -321,3 +331,5 @@ private val lightScrim = android.graphics.Color.argb(0xe6, 0xFF, 0xFF, 0xFF)
  * https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:activity/activity/src/main/java/androidx/activity/EdgeToEdge.kt;l=40-44;drc=27e7d52e8604a080133e8b842db10c89b4482598
  */
 private val darkScrim = android.graphics.Color.argb(0x80, 0x1b, 0x1b, 0x1b)
+
+internal val AppBasePath = "spark://catalog"
```
SimonMarquis, line 335: I'm wondering if this `catalog` part is actually useful for something?

catalog/src/main/kotlin/com/adevinta/spark/catalog/model/Components.kt
```diff
@@ -91,8 +91,34 @@ public data class Component(
 private var nextId: Int = 1
```
SimonMarquis, line 91: Just wondering, what are these id used for? if it is to reference unique things, we should use String keys. Also, I think these 2 lines could be removed now.

catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
soulcramer: Done in bc1dcd18a45a737e2f2a4245eb54f30529e1a4e1

catalog/src/main/kotlin/com/adevinta/spark/catalog/model/Components.kt
soulcramer: I removed and used string instead of int for the ids in a6915cfe739196b1b581040379926f45e3b887aa

catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
soulcramer: Done in 7b20af4f02332c94d5dab6704f845e144ae35eb8

catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApp.kt
soulcramer: Done in 7b20af4f02332c94d5dab6704f845e144ae35eb8

PR #1485, opened 2025-02-27
chore: cleanup Gradle Version Catalog

SimonMarquis:
## 📋 Changes

- delete unused entries
- inline versions if they are only used once
- add missing BoM for Coil
- space unrelated entries
- some minor renames to respect conventions


soulcramer (approved):
Thanks!

PR #1489, opened 2025-03-05
feat(catalog): Add french string res

soulcramer:
## 📋 Changes

Add generated translations from english to french for the catalog app ressources.
Add a few new a11y content strings to make the app more navigable with a screen reader.

## 🤔 Context

When testing a11y with a screen reader in Android we would get english text read as french which is not ideal.

## ✅ Checklist

### DIFF
--- a/catalog/src/main/res/values-fr/strings.xml
+++ b/catalog/src/main/res/values-fr/strings.xml
@@ -1,0 +1,143 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+  Copyright (c) 2025 Adevinta
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy
+  of this software and associated documentation files (the "Software"), to deal
+  in the Software without restriction, including without limitation the rights
+  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+  copies of the Software, and to permit persons to whom the Software is
+  furnished to do so, subject to the following conditions:
+
+  The above copyright notice and this permission notice shall be included in all
+  copies or substantial portions of the Software.
+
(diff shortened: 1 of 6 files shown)


catalog/src/main/res/values-fr/strings.xml
```diff
@@ -0,0 +1,143 @@
...
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+  SOFTWARE.
+  -->
+<resources xmlns:tools="http://schemas.android.com/tools" tools:locale="fr">
+    <string name="app_name">Spark</string>
+
+    <string name="brand">Sélectionner une marque</string>
+    <string name="pro">Utiliser le thème Pro</string>
+
+    <string name="example_tokens_elevation_label">Valeur d’élévation : %d</string>
```
SimonMarquis, line 30: This file is using curly quotes instead of straight quotes, which iirc are not correctly handled by talkback

soulcramer: I just tested it and there's no issues with the curly quote in french with talkback 🤔
I even verified it with using the english version of the app with the screen reader in french it read it like with a french accent and it has no problem en both talckback and the app is in english too

soulcramer: Maybe there was a fix in a previous version of Talkback, I'll look it up

soulcramer: There's nothing related to an issue on curly quote in the Talkback changelog and I couldn't find any existing issue in the Android Issue Tracker related to it 😬

Issue #1495, opened 2025-03-13
[Accessibility] Improve Text field screen reader behavior


Sildian:
# Description of the issue

The current screen reader behavior for Text field doesn't match the design requirements. Currently, only the value is read when it is present and the user has to swipe to read the label and the other information such as the helper text or the error message.

# Requirements

## Mobile screenreader
When I use a mobile screenreader (Talkback, VoiceOver) and I swipe to focus to an a FormField:
I hear a concise & sort description that conveys the content of the FormField elements: Label, if it’s mandatory, what can they do with the element, the helptext message.

Source : https://zeroheight.com/1186e1705/p/423a01-form-field

PR #1496, opened 2025-03-14
feat(a11y): Improve TextField screen reader behavior

Sildian:
## 📋 Changes

TextField screen reader behavior is improved to match the design requirements.
- Label: The label is now read by Talkback when the user focus on the field. There is no need to swipe anymore.
- Counter: When the counter is read by Talkback, its meaning is now obvious to the user.
- State message: The state status is now read by Talkback.

## 🤔 Context

Closes issue #1495 

When I use a mobile screenreader (Talkback, VoiceOver) and I swipe to focus to an a FormField:
I hear a concise & sort description that conveys the content of the FormField elements: Label, if it’s mandatory, what can they do with the element, the helptext message.

Source : https://zeroheight.com/1186e1705/p/423a01-form-field

## ✅ Checklist

## 📸 Screenshots

None

## 🗒️ Other info

None


spark/src/main/kotlin/com/adevinta/spark/components/textfields/SparkTextField.kt
```diff
@@ -442,16 +434,66 @@ private fun supportText(
...
+        append(label)
+        if (required) {
+            append(stringResource(id = R.string.spark_textfield_content_description_break))
+            append(stringResource(id = R.string.spark_textfield_mandatory_content_description))
+        }
+    }
+
+@Composable
+private fun computeStateMessageContentDescription(
+    state: TextFieldState?,
```
SimonMarquis, line 480: Do we actually need this to be nullable?

spark/src/main/kotlin/com/adevinta/spark/components/textfields/SparkTextField.kt
```diff
@@ -442,16 +434,66 @@ private fun supportText(
...
+    label: String?,
+): String =
+    buildString {
+        if (label == null) return@buildString
+        append(label)
+        if (required) {
+            append(stringResource(id = R.string.spark_textfield_content_description_break))
+            append(stringResource(id = R.string.spark_textfield_mandatory_content_description))
+        }
+    }
```
SimonMarquis, line 476: I don't think we should try to generate a sentence like this. Can we simply produce two sentences?

Here for instance it might produce this output: `Prénom\u00A0:\u00A0Champs requis`.
vs `Prénom\nChamps requis`, TalkBack will probably read it the same way.

spark/src/main/kotlin/com/adevinta/spark/components/textfields/SparkTextField.kt
```diff
@@ -442,16 +434,66 @@ private fun supportText(
...
+        val stateStatusContentDescription = when (state) {
+            TextFieldState.Success -> stringResource(id = R.string.spark_textfield_state_success_content_description)
+            TextFieldState.Alert -> stringResource(id = R.string.spark_textfield_state_alert_content_description)
+            TextFieldState.Error -> stringResource(id = R.string.spark_textfield_state_error_content_description)
+        }
+        append(stateStatusContentDescription)
+        if (stateMessage != null) {
+            append(stringResource(id = R.string.spark_textfield_content_description_break))
+            append(stateMessage)
+        }
```
SimonMarquis, line 494: nit: same question for this one

spark/src/main/res/values-fr/strings.xml
```diff
@@ -63,8 +63,17 @@
...
+    <!--region TextField-->
+    <string name="spark_textfield_mandatory_content_description">Champs requis</string>
+    <plurals name="spark_textfield_counter_content_description" tools:ignore="UnusedQuantity">
+        <item quantity="one">%1$d caractère saisi sur %2$d maximum</item>
+        <item quantity="other">%1$d caractères saisis sur %2$d maximum</item>
+        <item quantity="many">%1$d de caractères saisis sur %2$d maximum</item>
+    </plurals>
+    <string name="spark_textfield_state_success_content_description">Le champ est valide</string>
+    <string name="spark_textfield_state_alert_content_description">Le champ comporte une alerte</string>
+    <string name="spark_textfield_state_error_content_description">Le champ comporte une erreur</string>
```
SimonMarquis, line 75: nit: comporte vs contient

spark/src/main/res/values/strings.xml
```diff
@@ -59,8 +59,16 @@
...
-    <string name="spark_textfield_content_description">Mandatory</string>
+    <!--region TextField-->
+    <string name="spark_textfield_mandatory_content_description">Mandatory</string>
+    <plurals name="spark_textfield_counter_content_description">
+        <item quantity="one">%1$d character entered out of %2$d maximum</item>
+        <item quantity="other">%1$d characters entered out of %2$d maximum</item>
+    </plurals>
+    <string name="spark_textfield_state_success_content_description">The field is valid</string>
+    <string name="spark_textfield_state_alert_content_description">The field has an alert</string>
+    <string name="spark_textfield_state_error_content_description">The field has an error</string>
```
SimonMarquis, line 70: nit: has vs contains

spark/src/main/res/values-de/strings.xml
```diff
@@ -56,7 +56,8 @@
     <!--endregion-->
 
     <!--region TextField-->
-    <string name="spark_textfield_content_description">Pflichtfeld</string>
+    <string name="spark_textfield_mandatory_content_description">Pflichtfeld</string>
+    <string name="spark_textfield_content_description_break">:\u00A0</string>
     <string name="spark_textfield_delete_a11y">Löschen</string>
```
SimonMarquis, line 61: nit: no german translation?
Question to @soulcramer: do you know if we still want to support other languages (en/de)?

spark/src/main/kotlin/com/adevinta/spark/components/textfields/SparkTextField.kt
```diff
@@ -442,16 +434,66 @@ private fun supportText(
...
     }
 } else {
     null
 }
 
+@Composable
+private fun computeLabelContentDescription(
+    required: Boolean,
+    label: String?,
```
SimonMarquis, line 467: nit: If the label is null, shouldn't we produce no String instead of an empty string? (related to the question below, whenchecking for null input)

spark/src/main/kotlin/com/adevinta/spark/components/textfields/SparkTextField.kt
```diff
@@ -442,16 +434,66 @@ private fun supportText(
...
-        )
+        val stateMessageContentDescriptionModifier = if (state != null) {
+            val stateMessageContentDescription = computeStateMessageContentDescription(
+                state = state,
+                stateMessage = stateMessage,
+            )
+            Modifier.semantics { contentDescription = stateMessageContentDescription }
+        } else {
+            Modifier
+        }
```
SimonMarquis, line 445: Q: Shouldn't this be moved inside the Row composable? Or does hoisting it there has no unexpected effects?

spark/src/main/res/values-de/strings.xml
soulcramer: I don't think we want them, Let's remove them in the next release

Sildian: I guess there is no need to make the translations in german then

spark/src/main/kotlin/com/adevinta/spark/components/textfields/SparkTextField.kt
Sildian: This is indeed the expected behavior

spark/src/main/kotlin/com/adevinta/spark/components/textfields/SparkTextField.kt
Sildian: It doesn't change the behavior but the meaning is probably more obvious if I move it just above the related Text composable.

spark/src/main/kotlin/com/adevinta/spark/components/textfields/SparkTextField.kt
```diff
@@ -440,18 +433,64 @@ private fun supportText(
...
 }
 
+@Composable
+private fun computeLabelContentDescription(
+    label: String,
+    required: Boolean,
+): String = buildString {
+    append(label)
+    if (required) {
+        append("\n")
```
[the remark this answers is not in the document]
Sildian: Indeed

Sildian:
@soulcramer everything is ready to be merged if it is ok to you

soulcramer (approved):
LGTM, I'll probably have to rework the TextField with M3 1.7

Sildian:
Thanks ! Looks like I'm not authorized to merge the PR, can you or @SimonMarquis do it please ?

PR #1507, opened 2025-03-25
fix: ensure Modifiers are applied only once when using conditional Modifiers

EliottLujan:
## 🤔 Context

When using conditional Modifiers extensions, such as `ifTrue`, `ifFalse`, `ifNull` or `ifNotNull`, if the condition was met, then the upstream Modifier was applied twice.

If we look at the implementation of `ifNotNull`

```
public inline fun <T : Any> Modifier.ifNotNull(value: T?, builder: Modifier.(T) -> Modifier): Modifier =
    if (value != null) then(this.builder(value)) else this
```

We can notice that the receiver Modifier is applied twice in the chain of Modifiers when the condition is met. Once when calling `then` (which is the equivalent of `this.then`) and once when calling `this.builder`

For most Modifiers, applying it twice would not change anything, but for some exceptions, such as the `rotate` Modifier, then applying twice would result in double the rotation, which is the unexpected behavior.

## 📋 Changes

- Ensure the upstream Modifier is only applied once
- Add screenshot tests for conditional Modifiers

## ✅ Checklist


SimonMarquis (dismissed):

spark-screenshot-testing/src/test/kotlin/com/adevinta/spark/tools/modifiers/ConditionsScreenshot.kt
```diff
@@ -0,0 +1,125 @@
...
+class ConditionsScreenshot {
+
+    @get:Rule
+    val paparazzi = paparazziRule(
+        renderingMode = SHRINK,
+        deviceConfig = DeviceConfig.PIXEL_C,
+    )
+
+    @Test
+    fun `test conditions`() = paparazzi.sparkSnapshot(drawBackground = true) {
```
SimonMarquis, line 57: Last nitpick, test functions do not need to use the word "test”. Also, golden images are using the test method name in the filename, making it longer than necessary at the same time

spark-screenshot-testing/src/test/snapshots/images/com.adevinta.spark.textfields_DropdownScreenshot_expanded.png
SimonMarquis, line 1: Here we definitely spotted an issue that needs to be fixed.

spark-screenshot-testing/src/test/snapshots/images/com.adevinta.spark.tab_TabsScreenshot_enabled.png
SimonMarquis, line 1: Same here, the badges disappeared

spark-screenshot-testing/src/test/kotlin/com/adevinta/spark/tools/modifiers/ConditionsScreenshot.kt
EliottLujan: Done ✅

SimonMarquis:
@EliottLujan you are still commiting with another email account :p

soulcramer (approved):

PR #1514, opened 2025-03-27
fix: Fix Image Icon Size leading to crash

### LINKED ISSUE #1513: [Bug]: Wrong Image icon size leads to crash

- [x] I have searched the existing issues

1.1.0

### 🐛 What happened?

When displaying an `Image` without specifying a size, the icon displayed when the status is, for instance, Empty or Error, leads to a crash because of an out of range size. This bug was introduced within version 1.1.0.

### Relevant logcat output

```shell
java.lang.IllegalStateException: Size(996 x 2147483647) is out of range. Each dimension must be between 0 and 16777215.
                                                                                                    	at androidx.compose.ui.internal.InlineClassHelperKt.throwIllegalStateException(InlineClassHelper.kt:26)
                                                                                                    	at androidx.compose.ui.node.LookaheadCapablePlaceable.layout(LookaheadDelegate.kt:576)
                                                                                                    	at androidx.compose.ui.layout.MeasureScope.layout(MeasureScope.kt:54)
                                                                                                    	at androidx.compose.ui.layout.MeasureScope.layout$default(MeasureScope.kt:49)
                                                                                                    	at com.adevinta.spark.components.image.ImageKt.imageIconDynamicSize$lambda$10(Image.kt:289)
                                                                                                    	at com.adevinta.spark.components.image.ImageKt.$r8$lambda$U91Ov2eiuOHPdinRIUMD6iLfGgs(Unknown Source:0)
                                                                                                    	at com.adevinta.spark.components.image.ImageKt$$ExternalSyntheticLambda8.invoke(D8$$SyntheticClass:0)
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Sildian:
## 📋 Changes

Fix Image Icon Size leading to crash

## 🤔 Context

When displaying an `Image` without specifying a size, the icon displayed when the status is, for instance, Empty or Error, leads to a crash because of an out of range size. This bug was introduced within version 1.1.0.

**Example of stack trace:**

```
java.lang.IllegalStateException: Size(996 x 2147483647) is out of range. Each dimension must be between 0 and 16777215.
                                                                                                    	at androidx.compose.ui.internal.InlineClassHelperKt.throwIllegalStateException(InlineClassHelper.kt:26)
                                                                                                    	at androidx.compose.ui.node.LookaheadCapablePlaceable.layout(LookaheadDelegate.kt:576)
                                                                                                    	at androidx.compose.ui.layout.MeasureScope.layout(MeasureScope.kt:54)
                                                                                                    	at androidx.compose.ui.layout.MeasureScope.layout$default(MeasureScope.kt:49)
                                                                                                    	at com.adevinta.spark.components.image.ImageKt.imageIconDynamicSize$lambda$10(Image.kt:289)
                                                                                                    	at com.adevinta.spark.components.image.ImageKt.$r8$lambda$U91Ov2eiuOHPdinRIUMD6iLfGgs(Unknown Source:0)
                                                                                                    	at com.adevinta.spark.components.image.ImageKt$$ExternalSyntheticLambda8.invoke(D8$$SyntheticClass:0)
```

Closes #1513 

## ✅ Checklist

## 📸 Screenshots

None

## 🗒️ Other info

The bug was introduced within version 1.1.0, more specifically within [this PR](https://github.com/leboncoin/spark-android/commit/bff2ee960b89a9025199b311e15e517842eb8815). The size used to represent the bounding box where the icon is displayed was wrong, leading to a crash in the case where no size is provided to the image, because in this case no constraints are applied to define this bounding box. To fix this bug, we have to provide the calculated width and height to the bounding box.


soulcramer:
Hey!
Thanks for the proposal!
Do you have the screenshot test in lbc that show that this fix the issue?

I fear that not providing a fix size for the image will give us silent issues instead.
Currently the preview we have for this component looks like this:

We have 4 image that represent the states possible for this component with a fix size of `100.dp`
image

If  here we remove this fixed size we get this result:

The first Image will take the full constraints (even with the provided fix)
image

If we display them in a scrollable container we get this result a crash like you've mentioned but with the suggested fix we get a weird result:

The Empty and Error states will take the max size of the direction that has a bounded direction (width or height) but will take the min width/height required to be displayed (the icon in this case) however for the loading state, it won't be visible at all since it doesn't use an icon but its parent (the Image component) size to measure itself then both of them won't have a size.
image

Sildian:
@soulcramer , yes you're right ! Without a given size, the Loading state doesn't show anything. But isn't it normal in that case ? I mean, without a size the normal behavior of a Composable is to fit its content and when the Image is Loading there is no content, so there is nothing to show. I guess it's up to each developer to handle how to show an Image, with or without a size depending on the screen content. WDTY ?

soulcramer:
Since we now have a way to indicate that a Image component is missing a size we should be able to merge this! 😁

PR #1520, opened 2025-04-02
feat: show the entire expand/collapse animation in screenshot test

SimonMarquis:
The current implementation only showed half of the animation (and not the reverse part).
I also removed the unnecessary delay, and match the exact animation duration.


spark-screenshot-testing/src/test/snapshots/videos/com.adevinta.spark.icons_AnimatedIconsScreenshot_collapseExpand.png
SimonMarquis, line 1: 🐦🪽

soulcramer (approved):

PR #1535, opened 2025-04-24
feat(logger): add SparkLogger

soulcramer:
## 📋 Changes

Add SparkLogger fun interface to let lbc decide to crash/log/report on some crashable behaviors.

Note that the default implementation will make it crash.

## 🤔 Context

This change will make adding unsupported behaviours in currently used components safer as we can crash when they happen in debug builds and report a warning log in prod.

## ✅ Checklist


D3uf (approved):
LGTM 👍
I'm just wondering if adding a debug logging api along with the report one could be interesting today. We'll iterate over this point

PR #1551, opened 2025-05-13
chore(coil): Upgrade coil to 3.2 and use ImagePreviewHandler instead of custom transform

soulcramer:
## 📋 Changes

- Upgrade coil version to 3.2
- Change the import to use the coil v3 ones
- Use the new api for previews

## 🤔 Context

This version upgrade incorporate a lot of improvements (perfs, kmp for ex.)
We can now also use the new available apis to customize how images are rendered in previews which allow us to use screenshot testing and enhance the configurator experience.
Before these changes we were forced to change the image state to see the change of image type which was tedious.

## ✅ Checklist

## 📸 Screenshots

https://github.com/user-attachments/assets/f0c67806-5db4-450a-a8e2-3fb7a3240218


spark-screenshot-testing/src/test/kotlin/com/adevinta/spark/image/ImageScreenshot.kt
```diff
@@ -0,0 +1,266 @@
...
+            horizontalAlignment = Alignment.CenterHorizontally,
+            verticalArrangement = Arrangement.spacedBy(16.dp),
+            modifier = Modifier.padding(16.dp),
+        ) {
+            Text(
+                text = "Image Component",
+                style = SparkTheme.typography.headline1,
+            )
+
+            val states = listOf("Empty", "Loading", "Error", "Success")
```
SimonMarquis, line 120: Can we move this declaration to a local private enum class?
This would allow us to avoid hardcoding strings, and avoid future mistakes in the usage below.

soulcramer: Sure, I'll look if we can reuse the one we have for the Configurator

PR #1571, opened 2025-06-03
feat(catalog): remove brand theming options

soulcramer:
## 📋 Changes

Remove all reference to brands in the catalog app

## 🤔 Context

We don't really need these anymore.

## ✅ Checklist

## 📸 Screenshots

image

## 🗒️ Other info

Let's add the contrasted option in the picker later


catalog/src/main/kotlin/com/adevinta/spark/catalog/themes/Theme.kt
```diff
@@ -110,12 +102,7 @@ public enum class ThemeMode {
     Dark,
 }
 
-public enum class BrandMode(public val label: String) {
-    Leboncoin("Leboncoin (New Ui)"),
-    Kleinanzeigen("Kleinanzeigen"),
-    Milanuncios("Milanuncios"),
-    Subito("Subito"),
-}
+public enum class BrandMode(public val label: String)
```
SimonMarquis, line 105: I feel like we could also remove this class

SimonMarquis (approved):

PR #1581, opened 2025-06-11
feat(publish): Migrate publish from ossrh to Central Portal

soulcramer:
## 📋 Changes

Changes the GitHub Actions workflow to publish to the Central Portal instead of OSSRH. 
We also use Central publishing through the NMCP library from GradleUp  to stay with a simple publishing logic.
The necessary environment variables have also been updated to use the  Central Portal credentials instead of the OSSRH credentials.

## 🤔 Context

See https://central.sonatype.org/news/20250326_ossrh_sunset/

## ✅ Checklist

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,9 +2,10 @@
 
 ## Pre-release
 
-> This porcess will apply for the versions starting 1.4.0.
+> [!NOTE]
+> This process will apply for the versions starting 1.4.0.
 
-Before each release, we will publish one or more alpha version and publish pre-releases on [Github Release page](https://github.com/leboncoin/spark-android/releases) with the changes
+Before each release, we will publish one or more alpha versions and publish pre-releases on [Github Release page](https://github.com/leboncoin/spark-android/releases).
 
 Consumers will be able to test new features and report breaking changes & bugs that can be fixed before it’s considered for a stable release.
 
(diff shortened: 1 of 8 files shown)


build-logic/src/main/kotlin/com/adevinta/spark/SparkPublishingPlugin.kt
```diff
@@ -54,17 +57,14 @@ internal class SparkPublishingPlugin : Plugin<Project> {
...
-                }
-            }
+        }
+    }
+
+    private fun Project.configureNMCP() = configure<NmcpExtension> {
+        centralPortal {
+            username = System.getenv("CENTRAL_PORTAL_USERNAME")
+            password = System.getenv("CENTRAL_PORTAL_PASSWORD")
+            publishingType = "USER_MANAGED"
```
SimonMarquis, line 67: We could migrate to the automated publishing to simplify this a bit more now that we validated the new plugin.

RELEASING.md
```diff
@@ -2,25 +2,26 @@
 
 ## Pre-release
 
-Before each release, we will publish a Snapshot version of it and use it to create a pre-release on [Github Release page](https://github.com/leboncoin/spark-android/releases)
+> This porcess will apply for the versions starting 1.4.0.
```
SimonMarquis, line 5: Technically starting with 1.3.0 instead?

SimonMarquis (approved):

PR #1610, opened 2025-07-02
feat(build): update publishing configuration from automatic to user managed

soulcramer:
We get too much timeout with the automatic publishing type configuration so I'm switching it back to user


SimonMarquis:
FYI: for the timeout, I might open a new PR to set `publishingTimeout = Duration.ZERO` to not wait for full publication after this PR:
- https://github.com/leboncoin/spark-android/pull/1612

PR #1612, opened 2025-07-02
chore(ci): migrate to single aggregation publication

SimonMarquis:
## 📋 Changes

Migrate from multiple nmcp publications to a single aggregation nmcp publication.

This should prevent accumulating timeouts on each publication task.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: ./.github/actions/setup-gradle-properties
 
       - name: '📦 Publish to Central Portal'
-        run: ./gradlew publishAllPublicationsToCentralPortal --no-configuration-cache
+        run: ./gradlew publishAggregationToCentralPortal --no-configuration-cache
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CENTRAL_PORTAL_USERNAME: ${{ secrets.CENTRAL_PORTAL_USERNAME }}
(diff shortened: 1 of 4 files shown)


soulcramer (approved):

.github/workflows/publish.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - uses: ./.github/actions/setup-gradle-properties
 
       - name: '📦 Publish to Central Portal'
-        run: ./gradlew publishAllPublicationsToCentralPortal --no-configuration-cache
+        run: ./gradlew publishAggregationToCentralPortal --no-configuration-cache
```
martinbonnin, line 19: Small nit: this should work with the configuration cache 🎉 

Not that it matters much in the grand scheme of things but this was generally painful to implement so I'd thought I'd mention it 😄

SimonMarquis: IIRC Dokka was not happy with configuration cache, and since it is responsible for creating the javadoc, I feel like configuration cache should not be restored yet for this step 🤔

SimonMarquis: I looked into dokkatoo a long time ago, and wanted to migrate.
But the recent Dokka 2 might be CC compatible, but we need to correctly migrate first
- https://github.com/leboncoin/spark-android/pull/1406

martinbonnin: Yea, I think that was the plan but I haven't followed too much lately either...

PR #1648, opened 2025-07-30
feat(catalog): Add CatalogApplication and strict mode policy

soulcramer:
## 📋 Changes

Add new custom application class to set thread and vm  strict mode policies 

## 🤔 Context

Enable strict mode policies so that we may catch (or not 🤷) infractions earlier


catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApplication.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    override fun onCreate() {
+        super.onCreate()
+
+        setStrictModePolicy()
+    }
+
+    /**
+     * Return true if the application is debuggable.
+     */
```
SimonMarquis, line 40: self-explanatory comment 🙈

catalog/src/main/kotlin/com/adevinta/spark/catalog/CatalogApplication.kt
```diff
@@ -0,0 +1,59 @@
...
+    private fun isDebuggable(): Boolean = 0 != applicationInfo.flags and ApplicationInfo.FLAG_DEBUGGABLE
+
+    /**
+     * Set a thread & vm policy that detects all potential problems on the main thread, such as network
+     * and disk access.
+     *
+     * If a problem is found, the offending call will be logged.
+     */
```
SimonMarquis, line 48: Same, also, this comment is technically wrong as this does not only detect problems on the main thread.

PR #1719, opened 2025-10-15
fix(Scaffold): apply background token as background of Scaffold

EliottLujan:
## 📋 Changes

`Scaffold` now uses the `background` token as its background

## 🤔 Context

Until now `Scaffold` has been using the `surface` token as its background. We did not notice because with the part theme, the `surface` and `background` tokens have the same value

With the pro theme, those tokens have different values, highlighting that we've been using `surface` as the `Scaffold` background instead of `background`

Not that the `ModalScaffold` and `Dialogs` must still use the `surface` token as their background.

## ✅ Checklist

## 📸 Screenshots

Here is a before & after of a screen that uses the pro theme

| Before | After |
| ----- | ----- |
| Screenshot_20251015_182200 | Screenshot_20251015_171438 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,13 @@
 
 _2025-10-03_
 
-Revert "chore(deps): bump androidx.core:core-ktx from 1.16.0 to 1.17.0 (#1662)"
+### Spark
 
-The version 1.17.0 require consumers to upgrade their compile sdk version to api 36 but this breaks the unit test for modules where a roboeletric & paparazzi test is present.
+#### Dependency Updates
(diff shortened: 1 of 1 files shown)


SimonMarquis:
I see 2 issues with this change:
- the secondary button outline becomes almost invisible
- the screen background is not the same as the action bar

Are these on purpose, or are we discovering new issues that need to be addressed?

soulcramer (approved):
Thanks for the fix!
I'm looking at how the catalog app render and if it's okay I'll approve

soulcramer:
ah I forgot about it but can you add this change in the [main/CHANGELOG.md ](https://github.com/leboncoin/spark-android/blob/main/CHANGELOG.md) 🫣

PR #1814, opened 2026-01-21
chore(lint): rename issue id MaterialComposableUsageDetector

SimonMarquis:
## 📋 Changes

Rename lint issue ID from `MaterialComposableUsageDetector` to `MaterialComposableHasSparkReplacement` in order to avoid using the suffix Detector but most importantly to precisely define the reason of this detector. The original name was not meaningful.

## 🗒️ Other info

Suppressions will have to be update as well on the consumer side as there is no way to create aliases.


soulcramer (approved):
LGTM! I'll create a review to update the suppressions 👍

PR #1872, opened 2026-03-05
fix(docs): remove links to private components and removed screenshots

soulcramer:
## 🤔 Context

Using the feedbacks from https://github.com/leboncoin/spark-android/pull/1378 with linkchecker I fixed the links that were referencing private code and screenshots that were removed

### DIFF
--- a/spark/src/main/kotlin/com/adevinta/spark/components/iconbuttons/IconButton.kt
+++ b/spark/src/main/kotlin/com/adevinta/spark/components/iconbuttons/IconButton.kt
@@ -58,7 +58,7 @@ import com.adevinta.spark.tools.modifiers.sparkUsageOverlay
  * compact button is required, such as in a toolbar or image list.
  *
  * @param icon a content to be drawn inside the IconButton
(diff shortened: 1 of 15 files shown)


spark/src/main/kotlin/com/adevinta/spark/components/bottomsheet/BottomSheet.kt
```diff
@@ -82,7 +82,7 @@ import kotlinx.coroutines.launch
  *
  * If you want to have immersive BottomSheet, you can set contentTopPadding = 0.dp,
  * Beware you need to set your content top padding yourself
- * to avoid content to be hidden by the handle at least [SheetDefaults.ContentTopPadding]
+ * to avoid content to be hidden by the handle at least SheetDefaults.ContentTopPadding
```
SimonMarquis: Why do we remove code links? This will impact the generated documentation and we will no longer be able to click it.
If it is because the name changed, then we should fix this.
If it is because the name is not resolved, we should add the proper import.

PR #1955, opened 2026-05-13
fix(ci): match the spark-icons artefact name that changed since migrating to kmp

soulcramer:
<!--
  Please remove sections wisely!
  And checkout the contribution docs at https://github.com/leboncoin/spark-android/blob/main/docs/contributing.md
-->

## 📋 Changes

<!-- Describe your changes in details -->
  - Add spark-icons.aar to upload-artifact glob alongside *-release.aar                                                                                                                                                                                                                                                                             
  - Update diffuse find pattern from spark-icons-release.aar to spark-icons*.aar 

## 🤔 Context

<!-- Why is this change required? What problem does it solve? -->
<!-- If it solves an issue, add the steps to reproduce it. -->
<!-- Closes #1234 -->
KMP artefacts don't have the release suffix causing the diffuse step to fail since we didn't download the artefact and matched nothing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,9 @@ jobs:
         uses: actions/upload-artifact@v7
         with:
           name: aar-release
-          path: '**/build/outputs/aar/*-release.aar'
+          path: |
+            **/build/outputs/aar/*-release.aar
+            **/build/outputs/aar/spark-icons.aar
       - name: 📦 Archive JVM JARs
         uses: actions/upload-artifact@v7
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -67,7 +67,9 @@ jobs:
         uses: actions/upload-artifact@v7
         with:
           name: aar-release
-          path: '**/build/outputs/aar/*-release.aar'
+          path: |
+            **/build/outputs/aar/*-release.aar
+            **/build/outputs/aar/spark-icons.aar
```
SimonMarquis: Why not *.aar? I don't think we build debug variants anyways

soulcramer: You're right! I had a hard time debugging these issues (a lot of it coming from previous build from before mirgating the icons to kmp 🤡)

PR #2006, opened 2026-06-03
fix(lint): remove unneeded Composable annotation from ComposeRedundantComposable rule

### LINKED ISSUE #558: ComposeRedundantComposable false positive for composable default-argument values
`ComposeRedundantComposable` only inspects the function body, not parameter default values. Creating false positive for code like : 

```kotlin
@Composable
fun defaultValueFromComposableFunction() = ...

@Composable
fun anyFun(
    colors: Colors = defaultValueFromComposableFunction(),
) {
    // nothing @Composable called here
}
```

Affect versions: `1.5.0`, `1.5.1`

soulcramer:
We're still blocked until the fix https://github.com/slackhq/compose-lints/issues/558 is not released


soulcramer:
Seems to be fixed in `1.5.4`!

PR #2033, opened 2026-06-17
fix(screenshot): update the circular meter golden screenshots

soulcramer:
I suspect that when the image is small enough the rasterization make it different every time we take a screenshot making the test flaky.
To resolve it we will use the error state to represent the image variant


SimonMarquis:
@soulcramer the issue is probably on the Metter with image compression/rasterization. Can we use something else in the Meter version with an image in the center?

PR #2045, opened 2026-06-19
chore: upgrade compile sdk version to 37

soulcramer:
## 🤔 Context

New versions for androidx libraries was updated from 36 to 37 forcing us to upgrade too if we want to upgrade these libraries


gradle/libs.versions.toml
SimonMarquis: @soulcramer  What about also upgrading the targetSdk for the catalog app?

soulcramer: I build it locally and it seems safe to upgrade it too

SimonMarquis:
Now you'll have to deal with Robolectric 😜

soulcramer:
Yep I saw that I'm on it 💀

soulcramer:
🥳

spark/src/test/kotlin/com/adevinta/spark/button/ButtonTest.kt
```diff
@@ -31,6 +31,8 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import kotlin.intArrayOf
```
SimonMarquis, line 35: this should be removed

spark/src/test/kotlin/com/adevinta/spark/placeholder/PlaceholderTest.kt
```diff
@@ -50,7 +50,7 @@ import org.robolectric.shadows.ShadowPixelCopy
         "androidx.loader.content",
     ],
     shadows = [ShadowPixelCopy::class],
-    sdk = [33],
+    sdk = [36],
```
SimonMarquis, line 53: same

spark/src/test/kotlin/com/adevinta/spark/stepper/StepperTest.kt
```diff
@@ -49,6 +49,7 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
```
SimonMarquis, line 52: same

PR #2091, opened 2026-07-23
feat(button): add a flag to map legacy to new buttons

soulcramer:
## 📋 Changes

Add a new flag to make old buttons looks like the new rebranded one
Add a mapping in each old buttons to correspond to one of the rebranded one depending on which design and intent is used

## 🤔 Context

This will enable us to check if the mapping is correct on all places where we use a button in lc and allow designers to give us feedback on this mapping so we can adjust it.

## ✅ Checklist

### DIFF
--- a/spark/src/test/kotlin/com/adevinta/spark/button/ButtonStyleMapperTest.kt
+++ b/spark/src/test/kotlin/com/adevinta/spark/button/ButtonStyleMapperTest.kt
@@ -1,0 +1,184 @@
+/*
+ * Copyright (c) 2026 Adevinta
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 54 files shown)


spark/src/test/kotlin/com/adevinta/spark/button/ButtonStyleMapperTest.kt
```diff
@@ -0,0 +1,184 @@
...
+package com.adevinta.spark.button
+
+import com.adevinta.spark.components.buttons.ButtonIntent
+import com.adevinta.spark.components.buttons.ButtonStyle
+import com.adevinta.spark.components.buttons.ButtonStyleMapper
+import com.adevinta.spark.components.buttons.ButtonVariant
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class ButtonStyleMapperTest {
```
EliottLujan, line 31: nit: this test class would be a good candidate for ParameterizedTests 😄

spark/src/main/kotlin/com/adevinta/spark/SparkFeatureFlag.kt
```diff
@@ -30,10 +30,12 @@ package com.adevinta.spark
...
  * @property isContainingActivityEdgeToEdge
  * @property useRebrandedShapes Use new button, chips, tags and textfield shapes.
+ * @property useRebrandedButtons When true, old intent+style combinations resolve to new semantic button variants.
  */
 public data class SparkFeatureFlag(
     val useSparkTokensHighlighter: Boolean = false,
     val useSparkComponentsHighlighter: Boolean = false,
     val isContainingActivityEdgeToEdge: Boolean = false,
     val useRebrandedShapes: Boolean = false,
+    val useRebrandedButtons: Boolean = false,
```
EliottLujan, line 40: nit: any reason to have a different default value than `ThemeProperties`?

EliottLujan (dismissed):

spark/src/main/kotlin/com/adevinta/spark/SparkFeatureFlag.kt
soulcramer: Yep, `ThemeProperties` is used the equivalent of the flags we use in lbc so we ant to have it enabled by default for testing purpose and `SparkFeatureFlags` is the one we use internally in the library and since its not prod approved yet I've set it as disabled by default in case a dev use `SparkTheme` directly instead of the wrapper we use in lbc

spark/src/test/kotlin/com/adevinta/spark/button/ButtonStyleMapperTest.kt
soulcramer: Imo I would like to keep it as-is, the tests are short and each name reads as a spec. Parameterized tests will save lines but we  will lose that readability and we might have to adjust the spec based on feedback from the designers

PR #1208, opened 2024-07-08
fix: Divider designer comments

kazaky:
## 📋 Changes
- Fix empty space inside Divider when label is empty
- Fix Horizontal Divider disappears, when label is too big

## Summary by CodeRabbit

- **New Features**
  - Introduced new layout management for dividers using `ConstraintLayout`, enhancing alignment and styling.
  
- **Bug Fixes**
  - Improved text wrapping for divider labels to handle overflow better.

- **Dependencies**
  - Added support for `androidx.constraintlayout` and `androidx.constraintlayoutCompose` libraries.

PR #2166, opened 2026-08-14
chore(divider): remove ConstraintLayout dependency

soulcramer:
## 📋 Changes

- Replaces the ConstraintLayout-based label layout in `Divider` with `Row`/`Column` + weight. 
- The alignment-side line is fixed at 40.dp; the other side uses `weight(1f)` with a 40.dp minimum; the label uses `weight(1f, fill = false)`. Public API, enums, and no-label paths are unchanged. 
- Removes the `constraintlayout` dependency from `spark/build.gradle.kts` and its transitive View-system tail (appcompat, fragment, drawerlayout, viewpager, emoji2, ...) from the runtime classpath. 

Adds two new Paparazzi cases to `DividerScreenshot` for the empty-label and oversized-label edge cases that first motivated the ConstraintLayout layout in #1208.

## 🤔 Context

`constraintlayout` is the last Android-only dependency in the `spark` module. Removing it is a prerequisite for the KMP extraction tracked in #1928.

## ✅  Checklist


spark-screenshot-testing/src/test/snapshots/images/com.adevinta.spark.divider_DividerScreenshot_divider.png
soulcramer, line 1: The divider seems to not extend like we want...

